PR #114, opened 2023-07-10
Upload 300/supp api swagger

cfarmer-fearless:
Moved openapi spec content to main openapi file.  Also cleaned up some other routes and added some response types.  Deleted spec file for supplemental API function app as it's combined with the main openapi file.


swagger/openapi.yml
```diff
@@ -0,0 +1,807 @@
...
+paths:
+  /oauth:
+    get:
+      summary: This GET request goes to SAMS to validate a supplied token by the client.
+      description:
+      security:
+        - bearerAuth: []
+      responses:
+        '200':
+          description: JSON object describing the proveded auth token.
```
dancing-goats, line 25: nit: provided*

dancing-goats (changes requested):

swagger/openapi.yml
cfarmer-fearless: Done

dancing-goats (approved):

PR #116, opened 2023-07-11
moving use-case definitions in a folder

cosmycx:


cfarmer-fearless:
Looks like the unit test workflow needs to be updated to use SaaS runner instead of self-hosted runner.

cfarmer-fearless:
Just fyi, this will break the CD pipeline for the tus hooks.  This file need to be updated: 
https://github.com/cdcent/data-exchange-upload-devops/blob/main/.github/workflows/tus-pre-create-hook-cd-dev.yml

candace-campbell (approved):

PR #117, opened 2023-07-11
Added workflows to deploy to staging

ccampbell-fearless:
[Upload-293](https://cdc-dexops.atlassian.net/browse/UPLOAD-293)

Related PR: https://github.com/cdcent/data-exchange-upload-devops/pull/3

### DIFF
--- a/.github/workflows/deploy-rc-staging.yml
+++ b/.github/workflows/deploy-rc-staging.yml
@@ -1,4 +1,4 @@
-name: "Deploy Function App Release Candidate - Staging"
+name: "Deploy Release Candidate - Staging"
 
 on:
   workflow_dispatch:
@@ -11,5 +11,5 @@ jobs:
     uses: ./.github/workflows/remote-cd-trigger-template.yml
     with:
       WORKFLOW: function-app-rc-matrix.yml
-      REF: ${{ github.ref }}
+      REF: ${{ github.ref }}  # Resolves to the tag that is pushed
     secrets: inherit
--- a/.github/workflows/deploy-rc-staging.yml
+++ b/.github/workflows/deploy-rc-staging.yml
@@ -1,17 +1,15 @@
-name: "Supplemental API Function App Remote CD Staging"
+name: "Deploy Function App Release Candidate - Staging"
 
 on:
   workflow_dispatch:
   push:
     tags:
       - '*'
-    paths:
-      - supplemental-api-function-app/**
 
 jobs:
   staging:
     uses: ./.github/workflows/remote-cd-trigger-template.yml
     with:
-      WORKFLOW: supplemental-api-function-app-cd-staging.yml
+      WORKFLOW: function-app-rc-matrix.yml
       REF: ${{ github.ref }}
     secrets: inherit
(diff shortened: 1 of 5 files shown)


cfarmer-fearless (changes requested):

cfarmer-fearless (changes requested):
Small changes but looks good overall!

.github/workflows/deploy-rc-staging.yml
```diff
@@ -0,0 +1,15 @@
...
+  push:
+    tags:
+      - '*'
+
+jobs:
+  staging:
+    uses: ./.github/workflows/remote-cd-trigger-template.yml
+    with:
+      WORKFLOW: function-app-rc-matrix.yml
+      REF: ${{ github.ref }}
```
cfarmer-fearless, line 14: Add a comment that this will assume the value of the tag that was pushed.

.github/workflows/remote-cd-trigger-template.yml
```diff
@@ -7,6 +7,10 @@ on:
         type: string
         description: "Workflow yml file that should be triggered."
         required: true
+      REF: 
+        type: string
+        description: "Github reference"
```
cfarmer-fearless, line 12: This description can have a bit more detail.  This value should be the tag or branch that the workflow is running on.

.github/workflows/remote-cd-trigger-template.yml
```diff
@@ -7,6 +7,10 @@ on:
...
+        required: true
```
cfarmer-fearless, line 13: Let's set this required to false and default to `main` so we don't need to change the dev workflows.

cfarmer-fearless (changes requested):

.github/workflows/remote-cd-trigger-template.yml
```diff
@@ -34,5 +38,5 @@ jobs:
                   owner: 'cdcent',
                   repo: 'data-exchange-upload-devops',
                   workflow_id: '${{ inputs.WORKFLOW }}',
-                  ref: 'main'
```
cfarmer-fearless, line 37: This needs to stay hard-coded to main.  This ref is referring to the branch on `data-exchange-upload-devops` that is to be checked out, not the tag on `data-exchange-upload`, which should always be `main`.

.github/workflows/remote-cd-trigger-template.yml
```diff
@@ -34,5 +38,5 @@ jobs:
...
+                  ref: '${{ inputs.REF }}'
```
cfarmer-fearless, line 41: Here, the ref will be the tag.  We need to pass it via the `inputs` object.  Should look something like this:
```
ref: 'main',
inputs: {
    DATA_EXCHANGE_UPLOAD_TAG: ${{ input.REF }}
},
```

cfarmer-fearless (changes requested):

cfarmer-fearless (changes requested):
Don't need to set `REF` to `main` in the dev workflows since you default to main.  Looks good otherwise!

cfarmer-fearless (approved):
Great!

PR #123, opened 2023-07-20
Added quotes to git reference and input value

candace-campbell:


.github/workflows/deploy-rc-staging.yml
```diff
@@ -11,5 +11,5 @@ jobs:
     uses: ./.github/workflows/remote-cd-trigger-template.yml
     with:
       WORKFLOW: function-app-rc.yml
-      REF: ${{ github.ref_name }}  # Resolves to the tag that is pushed
+      REF: '${{ github.ref_name }}'  # Resolves to the tag that is pushed
```
cfarmer-fearless, line 14: I think the quotes need to go around the `inputs.REF` instead of here.

cfarmer-fearless (changes requested):

.github/workflows/deploy-rc-staging.yml
candace-campbell: They're in both places but the other one was merged in yesterday so it doesn't show up in the diff

cfarmer-fearless: Ah I see it now

cfarmer-fearless (approved):

PR #128, opened 2023-07-24
copy of metadata to edav hot fix

cosmycx:


rohitpanwar (approved):
LGTM

mkrystof (approved):
Thanks for changing it to do the upload and metadata all in one call.  Looks good to me now, thank you.

PR #133, opened 2023-08-07
UPLOAD-462 bulk upload using root container name from config for edav storage

cosmycx:


cosmycx:
On hold for VNET migration

mkrystof (commented):
To make sure I understand -- the purpose of this is to have the ability to have a single upload container in EDAV and then all uploads are under this container name as subfolders?  Is this going to be be turned off for now?  Because if it's not this will definitely break NDLP as they have EDAV triggers for blob created events on their container names.

cosmycx:
@mkrystof yes we'll need a root container in Edav as more sources go to this Edav storage account: e.g. fhir, router, buzz, etc.. so each needs to have their own root.
yes that's why we did not pushed it yet, it needs coordination with current prod. use-case.
we'd like to not onboard more prod. use-cases before this goes in for obvious reasons. 
we need to discuss this with @rohitpanwar and @anandCD405 . let's do that soon. I'm open to ideas. we can discuss internally.

mkrystof:
> @mkrystof yes we'll need a root container in Edav as more sources go to this Edav storage account: e.g. fhir, router, buzz, etc.. so each needs to have their own root. yes that's why we did not pushed it yet, it needs coordination with current prod. use-case. we'd like to not onboard more prod. use-cases before this goes in for obvious reasons. we need to discuss this with @rohitpanwar and @anandCD405 . let's do that soon. I'm open to ideas. we can discuss internally.

If we add the event hub notifications feature request, where programs are notified of uploads by topic then this would be moot and totally fine.  With that feature NDLP and other programs wouldn't need to be tied to EDAV blob created events and the file location within EDAV wouldn't matter.  :)

cosmycx:
I like the idea, there would be no breaking program changes except they need to move to event hub which is what hey want, plus we'd be onboarding all of the above, we'd need to make the timeframe work and concurrence @rmharrison

rohitpanwar (approved):
LGTM

PR #155, opened 2023-09-18
upload-498 implement unit test for supplemental api

USZ7:
- Implement unit test for supplemental api using TestNG

### DIFF
--- a/supplemental-api-function-app/src/test/kotlin/test/HealthCheckFunctionTest.kt
+++ b/supplemental-api-function-app/src/test/kotlin/test/HealthCheckFunctionTest.kt
@@ -1,45 +1,41 @@
 package test
 
+import com.azure.cosmos.CosmosClient
 import com.microsoft.azure.functions.ExecutionContext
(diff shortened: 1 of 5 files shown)


supplemental-api-function-app/src/test/kotlin/test/HealthCheckFunctionTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        val healthCheckFunction = HealthCheckFunction()
+
+        // Call healthCheckFunction.run() with the mock objects
+        val response = healthCheckFunction.run(request, context)
+
+        // Assert the response as needed for the success scenario
+
+        assert(response == HttpStatus.OK)
+
+    }
```
rohitpanwar, line 40: Can we add 2 scenarios here?   1. Test 200 Ok back 2. Test 500 error back

rohitpanwar (changes requested):

rohitpanwar (approved):

PR #159, opened 2023-09-22
Initial version of ndlp apl historical data config

mkrystof:


cosmycx:
@mkrystof  definition file is empty, I think it needs at least the file name and some more meta would be ok.

mkrystof:
> @mkrystof definition file is empty, I think it needs at least the file name and some more meta would be ok.

I think it's okay there are no "additional" required metadata fields in the definition.  The definition file isn't even required, but if provided, it gives the additional meta fields besides the ones required - filename, destination_id, and meta_ext_event.  So, unless something has changed I think the way it is now is okay for now.  

The end user may add some additional required metadata fields, but wanted it wide-open for now as they decide what they want.

cosmycx (approved):

PR #171, opened 2023-10-25
Upload 641/destination id endpoint

cfarmer-fearless:
Adds an endpoint at /api/destination for returning all allowed destinations.


supplemental-api-function-app/src/main/java/gov/cdc/ocio/supplementalapi/functions/DestinationIdFunction.java
```diff
@@ -0,0 +1,48 @@
...
+
+        if (destinations == null) {
+            return request.createResponseBuilder(HttpStatus.INTERNAL_SERVER_ERROR).build();
+        }
+
+        List<String> destStrs = Arrays.stream(destinations).map(destination -> destination.destinationId).collect(Collectors.toList());
+
+        return request.createResponseBuilder(HttpStatus.OK).header("Content-Type", "application/json").body(destStrs).build();
+    }
+}
```
rohitpanwar, line 48: could you please add unit test for the same

supplemental-api-function-app/src/main/java/gov/cdc/ocio/supplementalapi/model/Destination.java
USZ7, line 1: Why use java instead of Kotlin? Since the function is in Kotlin.

cfarmer-fearless: Thanks for calling this out!  I tried kotlin at first but was running into issues importing the azure sdk classes.  I thought there was an incompatibility with kotlin, but I think the real issue was I was including them incorrectly in the build.gradle.  That's fixed and I was able to refactor to Kotlin!

USZ7: Looks good now.

USZ7 (approved):
lgtm!

supplemental-api-function-app/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# For local development only.
+systemProp.subscription=OCIO-EDEDEV-C1
+systemProp.resourceGroup=ocio-ede-dev-moderate-hl7-rg
+systemProp.appName=ocio-ede-dev-supplemental-api
```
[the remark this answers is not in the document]
cfarmer-fearless: Went ahead and removed this file for now and added steps in the readme for running locally.

supplemental-api-function-app/src/main/java/gov/cdc/ocio/supplementalapi/functions/DestinationIdFunction.java
cfarmer-fearless: Added two unit tests: one for happy path, and one with an empty json file.

PR #173, opened 2023-10-29
upload-555 integrate app insight

USZ7:
- Integrate App Insight SDK for tracing support.

## Reference
- https://learn.microsoft.com/en-us/azure/azure-monitor/app/worker-service

### DIFF
--- a/BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
+++ b/BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
@@ -13,6 +13,8 @@ using Azure.Messaging.EventHubs.Producer;
 using Azure.Messaging.EventHubs;
 using Newtonsoft.Json;
 using BulkFileUploadFunctionApp.Utils;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
 
 namespace BulkFileUploadFunctionApp
 {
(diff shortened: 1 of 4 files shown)


BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
```diff
@@ -94,6 +111,8 @@ public async Task Run([EventHubTrigger("%AzureEventHubName%", Connection = "Azur
 
                 await ProcessBlobCreatedEvent(blobCreatedEvent?.Data?.Url);
 
+                metric.TrackValue(1);
```
emmanuel-apau, line 114: what does this do?

USZ7: After processing each blob increment the count by 1 for each blob processed.

emmanuel-apau: ok could you send me a link via teams so i can see this an action

emmanuel-apau:
Is there a wrapper that can be used vs the client specific commands?

emmanuel-apau:
did you look at this page? https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable?tabs=aspnetcore#modify-your-application

USZ7:
> Is there a wrapper that can be used vs the client specific commands?

I added wrapper class `TracingWrapper`

emmanuel-apau (approved):
lgtm. good work

PR #194, opened 2024-01-16
upload-748-788 remove sup and cosmos sync pipelines in CDCgov

USZ7:
- Remove sup and cosmos sync pipelines in CDCgov.


candace-campbell (approved):
LGTM

PR #195, opened 2024-01-18
upload-885 add feature flag for routing changes

USZ7:
- Add feature flag for routing changes.

### DIFF
--- a/BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
+++ b/BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
@@ -130,15 +130,15 @@ namespace BulkFileUploadFunctionApp
 
                 if (isRoutingEnabled)
                   {
-                    _logger.LogInformation($"Routing is Enabled. Bypassing routing for blob: {blobCreatedEvent?.Data?.Url}");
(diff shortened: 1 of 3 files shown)


BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
```diff
@@ -119,7 +128,17 @@ public async Task Run([EventHubTrigger("%AzureEventHubName%", Connection = "Azur
                 if (blobCreatedEvent == null)
                     throw new Exception("Unexpected data content of event; there should be at least one element in the array");
 
-                await ProcessBlobCreatedEvent(blobCreatedEvent?.Data?.Url);
+                if (isRoutingEnabled)
```
cfarmer-fearless, line 131: I think you want to flip the logic here.  If `isRoutingEnabled` is true, we don't want to bypass routing.

cfarmer-fearless (changes requested):

BulkFileUploadFunctionApp/BulkFileUploadFunction.cs
USZ7: done!

cfarmer-fearless (approved):

PR #196, opened 2024-01-18
Upload 785/upload trace start

cfarmer-fearless:
Introduces the post-create tus hook for facilitating processing status integration.  This hook is called when metadata is valid, and an upload is about to begin (i.e. right after the first post request is sent to the tus server).  The post-create hook will inform PS API of this event by first creating a trace.  This will be the global trace that follows the file through the entire dex process.  Finally, it starts a child span for capturing the tus resumable upload process, under the "dex-upload" stage name.  This span will be stopped in a later hook, when the upload is complete or terminated.

Testing steps:
You can run this python script from within the CDC network, providing dummy data with the following command: 
`python post_create_bin.py -m "{\"meta_destination_id\": \"1234\", \"meta_ext_event\": \"test_event\"}" -i "1234"`
This should output the trace ID.

You can take that trace ID and query the PS API by sending a GET request to the following endpoint:
https://ocio-ede-dev-pstatus-api.azurewebsites.net/api/trace/traceId/{traceId}

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,11 @@ tus/file-hooks/metadata-verify/pre_create_bin.spec
 
 tus/file-hooks/metadata-verify/pre-create-bin
 
+tus/file-hooks/post-create/build/
+tus/file-hooks/post-create/dist
+tus/file-hooks/post-create/post-create-bin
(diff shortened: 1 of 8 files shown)


mkrystof (approved):
lgtm!

PR #197, opened 2024-01-23
Upload 746/post hook to ps

rohitpanwar:


tus/file-hooks/upload-status/post_receive_bin.py
```diff
@@ -135,8 +120,8 @@ def main(argv):
             size = arg
         elif opt in ("-m", "--metadata"):
             metadata = arg
-    try:
-        post_receive(tus_id, int(offset), int(size), metadata)
+    try:        
+        asyncio.run(post_receive(tus_id, int(offset), int(size), metadata))
```
cfarmer-fearless, line 124: What does the `asyncio.run` do?

tus/file-hooks/upload-status/post_receive_bin.py
```diff
@@ -111,8 +82,22 @@ def post_receive(tguid, offset, size, metadata_json):
         # convert metadata json string to a dictionary
         metadata_json_dict = ast.literal_eval(metadata_json)
```
cfarmer-fearless, line 83: Why are we deserializing this json string just to re-serialize it on line 99 with the `json.dumps` function?

tus/file-hooks/upload-status/post_receive_bin.py
cfarmer-fearless: Future reference: async/await syntax for Python.  Shouldn't cause performance issues.

tus/file-hooks/upload-status/post_receive_bin.py
rohitpanwar: made the change to add up json strings instead

tus/file-hooks/upload-status/post_receive_bin.py
rohitpanwar: leaving this as is as we decided

cfarmer-fearless (approved):

PR #207, opened 2024-01-29
UPLOAD-942/remove call to metadata relay

cfarmer-fearless:
Simply removes the code for sending a file's metadata to NDLP.


cfarmer-fearless:
Manual test conducted verifying file made it to edav.

mkrystof (commented):
Looks good to me.  Should make it a little more performant too removing the unused extra step.

PR #222, opened 2024-02-06
Upload 966/report metadata verify failed ps api

cfarmer-fearless:
Expands on the pre-create hook to send a metadata-verify report to the PS API when there is a verification error.

Test steps for required metadata:
1. Navigate to the tus/file-hooks/metadata-verify directory
2. Create a `.env` file and add the `PS_API_URL` field and assign it the value of https://ocio-ede-dev-pstatus-api.azurewebsites.net
3. Run the python script with the following command: `python pre_create_bin.py --metadata "{\"filename\":\"test.txt\"}"`.  Notice that we are providing an empty metadata object here.
4. Verify that the script fails and prints that the destination ID and event type were not provided.  Also note the upload ID that was printed.
5. Query the PS API at https://https://ocio-ede-dev-pstatus-api.azurewebsites.net/api/report/json/uploadId/{uploadId} and provide the printed upload ID.  Verify that a report is returned with a message in the `issues` field of the response body saying that the destination and event type were not provided.
6. Query the PS API at https://ocio-ede-dev-pstatus-api.azurewebsites.net/api/trace/uploadId/{uploadId} providing the upload ID in the path.  Verify that a trace is returned with a single span for `dex-metadata-verify` and it has the status of complete.


tus/file-hooks/metadata-verify/pre_create_bin.py
```diff
@@ -1,124 +1,260 @@
...
+    return 'Found the following metadata validation errors: ' + ','.join(messages)
+
+
+def get_filename_from_metadata(meta_json):
+    filename_metadata_fields = ['filename', 'original_filename', 'meta_ext_filename']
+    filename = None
+
+    for field in meta_json:
+        if field in filename_metadata_fields:
+            filename = meta_json[field]
```
rohitpanwar, line 213: should iterate tru all the keys in json or filename_metadata_fields ? json will be much larger than 3 fields array

cfarmer-fearless: Feels like the same to me.  If we iterate through filename_metadata_fields, we need to search through the keys in the metadata json dictionary anyway.

cfarmer-fearless: Done

rohitpanwar (approved):

PR #227, opened 2024-02-12
Upload 983/ps api int tests

cfarmer-fearless:
New set of integration tests for the Upload API. This first set focuses on metadata verify and upload status with processing status integration. These tests are written in Kotlin and use TestNG. They can be run with gradle using gradle test. They require the following environment variables:

UPLOAD_URL
PS_API_URL
SAMS_USERNAME
SAMS_PASSWORD

These can be passed as JVM arguments (-D flags) or by creating a local.properties file in the root of the project. The local.properties file needs to look like this:

upload.url=https://apidev.cdc.gov
ps.api.url=https://ocio-ede-dev-pstatus-api.azurewebsites.net
sams.username=****
sams.password=****


rohitpanwar (approved):
LGTM

PR #240, opened 2024-02-27
Upload 924/v2 metadata hydration

cfarmer-fearless:
Adds hydration for v2 metadata.  Restructures the upload configs to be placed under a new V2 directory.  Also adds lists of metadata fields for each use case that supports default value, and v1 compatible field names.  

To test, upload a file to the dex testing destination and verify that the file in the dex and destination storage accounts have the following fields: 
- `sender_id`
- `jurisdiction`
- `data_producer_id`
- `received_filename`


upload-processor/BulkFileUploadFunctionApp/Services/UploadProcessingService.cs
```diff
@@ -545,6 +502,51 @@ private string GetFilenameSuffix(UploadConfig uploadConfig, DateTime dateTimeNow
...
+
+                tusInfoFile.MetaData.Add(field.FieldName, "");
+            }
+
+            // Add common fields and their values.
+            tusInfoFile.MetaData.Add("tus_tguid", tusInfoFile.ID); // TODO: verify this field can be replaced with upload_id only.
+            tusInfoFile.MetaData.Add("upload_id", tusInfoFile.ID);
+            tusInfoFile.MetaData.Add("trace_id", traceId);
+            tusInfoFile.MetaData.Add("span_id", spanId);
+            tusInfoFile.MetaData.Remove("filename");
```
USZ7, line 547: Just noticed you're removing the 'filename' metadata at the end. If that's intentional, a quick comment would help clarify why.

USZ7 (approved):
lgtm!

upload-processor/BulkFileUploadFunctionApp/Services/UploadProcessingService.cs
cfarmer-fearless: Good call, I added a comment

PR #246, opened 2024-02-29
upload-1017 integration tests using node typescript and playwright

USZ7:
-Integration tests using node typescript and playwright.


tests/integration/playwright/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      }
+    },
+  };
+
+  const upload = new tus.Upload(file, options);
+  upload.start();
+  return { uploadId, accessToken: loginResponse.accessToken };
+}
+
+export async function getTraceResponse(uploadId: string, accessToken: string): Promise<void> {
```
cfarmer-fearless, line 76: Don't need to pass an access token to PS API

tests/integration/playwright/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      const percentage = ((bytesUploaded / bytesTotal) * 100).toFixed(2);
+      console.log(bytesUploaded, bytesTotal, `${percentage}%`);
+    },
+    onSuccess() {
+      console.log("Upload finished:", url);
+
+      // Extract uploadId from the upload.url
+      const uploadIdPattern = /files\/([a-zA-Z0-9]+)/;
+      const matches = url.match(uploadIdPattern);
+      if (matches && matches[1]) {
```
cfarmer-fearless, line 59: Instead of doing this, use the `onUploadUrlAvailable()` to get the upload ID

tests/integration/playwright/index.ts
```diff
@@ -0,0 +1,91 @@
...
+        uploadId = matches[1];
+        console.log("Extracted uploadId:", uploadId);        
+
+        // Now make a GET request to the PS API with this uploadId
+        getTraceResponse(uploadId, loginResponse.accessToken);
```
cfarmer-fearless, line 64: Why do this here?

tests/integration/playwright/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      } else {
+        console.error("Could not extract uploadId from upload URL.");
+      }
+    },
+  };
+
+  const upload = new tus.Upload(file, options);
+  upload.start();
+  return { uploadId, accessToken: loginResponse.accessToken };
```
cfarmer-fearless, line 73: I don't think this `return` will wait until the upload is complete

tests/integration/playwright/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+
+
+export async function start(): Promise<{ uploadId: string, accessToken: string }> {
+  const [username, password, url, ps_url] = config.validateEnv();
+ 
+  const loginResponse: LoginResponse | null = await c.login(username, password, url); 
+  let uploadId = '';  
+  
+  if (loginResponse === null) {
```
cfarmer-fearless, line 17: Logging in and uploading a file should be totally decoupled.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+// Use test.describe to group your tests and use hooks like beforeAll
+test.describe('File Upload and Trace Response Flow', () => {
+  let uploadId: string;
+  let accessToken: string;
+
+  // Arrange
+  // Correctly use test.beforeAll at the describe level to perform setup actions
+  test.beforeAll(async () => {
+    // Perform setup, including logging in and uploading the file to get an upload ID
+    const result = await start();
```
cfarmer-fearless, line 13: The access token and upload ID should be decoupled.  Should be more like 
First: login and get access token
Second: Upload a file and get upload ID when complete

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    uploadId = result.uploadId;
+    accessToken = result.accessToken;
+  });
+
+  // Act & Assert 
+  test('upload and get trace response', async () => {
+    // Assert that we have an upload ID and an access Token
+    expect(uploadId).toBeTruthy();
```
cfarmer-fearless, line 21: You should expect it to equal an actual string value, not just truthy.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(accessToken).toBeTruthy();
+
+    // Act: Query the PS API with the obtained uploadId and accessToken
+    const traceResponse = await getTraceResponse(uploadId, accessToken);
```
cfarmer-fearless, line 25: I would like these API calls to use Playwright's API context instead of Axios. https://playwright.dev/java/docs/api-testing#introduction

cfarmer-fearless (changes requested):

tests/integration/playwright/client.ts
```diff
@@ -0,0 +1,39 @@
...
+    const params = new URLSearchParams({
+      username: username,
+      password: password,
+    });
+
+    try {
+            
+      const response = await axios.post(`${url}/oauth`, params);
+
+      console.log("Raw response data:", response.data);
```
cfarmer-fearless, line 21: Let's make sure to remove this log before it goes into main.  We don't want to be logging this data.  It's fine for debugging though

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  let accessToken: string;
+  let psApiUrl: string;
+
+  // Arrange
+  // Correctly use test.beforeAll at the describe level to perform setup actions
+  test.beforeAll(async ({ request }) => {
+    try {
+      accessToken = await loginAndGetToken();     
+      uploadId = await uploadFileAndGetId(accessToken);      
+      psApiUrl = `${config.validateEnv()[3]}/api/trace/uploadId/${uploadId}`;
```
cfarmer-fearless, line 18: I don't like having to get specific environment variables by indexing an array.  We should be able to get them by passing a key representing the variable's name.  In NodeJS, this is typically done with `process.env`.

tests/integration/playwright/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "playwright",
+  "version": "1.0.0",
+  "description": "",
+  "main": "dist/main.js",
+  "scripts": {
+    "build": "npx tsc",
+    "test": "npx playwright test"
+  },
+  "dependencies": {
```
cfarmer-fearless, line 10: Should there be a playwright dependency here?  See playwright installation docs: https://playwright.dev/docs/intro

tests/integration/playwright/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "axios": "^1.6.7",
```
cfarmer-fearless, line 11: Hoping we can remove `axios` as a dependency and just use [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch).

tests/integration/playwright/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "dotenv": "^16.4.5",
+    "express": "^4.17.1",
```
cfarmer-fearless, line 13: What do we need `express` for?

tests/integration/playwright/test/upload-test.spec.ts
cfarmer-fearless, line 18: Also, since tracing is currently broken on the PS API, let's validate data against the report endpoint instead.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    } catch (error) {
+      console.error("Setup failed:", error);
+      throw new Error("Failed to complete test setup.");
+    }
+  });
+
+  // Act & Assert 
+  test('upload and get trace response', async ({ request }) => {    
+
+    // Act: Query the PS API with the obtained uploadId and accessToken
```
cfarmer-fearless, line 29: Again, the PS API doesn't need an access token.  It can be removed from the request headers.  It's only needed for uploading a file.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+    });
+
+    // Parse the JSON response body
+    const responseBody = await response.json();
+    console.error("responseBody:", responseBody);
+
+    // Assert: Validate the data coming back from PS API equals an expected value
+    // Replace 'expectedProperty' and 'expectedValue' with actual data you expect
+    expect(responseBody).toHaveProperty('expectedProperty');
```
cfarmer-fearless, line 42: Shouldn't we have an actual expected value instead of just a string "expectedProperty"?  Seems like this code was just copy/pasted from somewhere.

tests/integration/playwright/upload.ts
```diff
@@ -0,0 +1,61 @@
+import fs from 'fs';
+import path from 'path';
+import * as tus from 'tus-js-client';
+import { v4 as uuidv4 } from 'uuid';
+import config from './config';
+
+export async function uploadFileAndGetId(accessToken) {
+  const [_, __, url] = config.validateEnv();
```
cfarmer-fearless, line 8: Since we only need the url here, let's just get it from the environment variables directly with `process.env`.

tests/integration/playwright/upload.ts
```diff
@@ -0,0 +1,61 @@
+import fs from 'fs';
+import path from 'path';
+import * as tus from 'tus-js-client';
+import { v4 as uuidv4 } from 'uuid';
+import config from './config';
+
+export async function uploadFileAndGetId(accessToken) {
```
cfarmer-fearless, line 7: Let's pass the filename here so we can upload different files.

tests/integration/playwright/upload.ts
```diff
@@ -0,0 +1,61 @@
...
+  const [_, __, url] = config.validateEnv();   
+  const filePath = path.resolve(__dirname,'..','upload-files', '10MB-test-file');   
+  const file = fs.createReadStream(filePath);
+
+  return new Promise((resolve, reject) => {
+    const options = {
+      endpoint: `${url}/upload`,
```
cfarmer-fearless, line 14: Let's pass the entire url and path instead of just the url so we can bypass APIM if we need to.

tests/integration/playwright/upload.ts
```diff
@@ -0,0 +1,61 @@
...
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+      },
+      metadata: {
```
cfarmer-fearless, line 18: Let's pass in this metadata so we have the ability to create tests based on certain metadata.  It may help to create helper functions to create metadata objects.

tests/integration/playwright/upload.ts
```diff
@@ -0,0 +1,61 @@
...
+      },
+    };
+
+    const upload = new tus.Upload(file, options);
+    upload.start();
+  });
+}
+
+function extractUploadId(uploadUrl) {
+    return new Promise((resolve, reject) => {
```
cfarmer-fearless, line 51: Does this need to return a promise?  Everything should be synchronous.

cfarmer-fearless (changes requested):
Still some work to do.  Please also create a README.md in the `playwright` directory with instructions on how to run these tests.

tests/integration/playwright/client.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface LoginResponse {
+  // Define the structure of your login response here
+  // Example:
+  access_token: string;
+  // Add other fields as necessary
+}
+
+const c = {
+  async login(username: string, password: string, url: string): Promise<LoginResponse | null> {
```
cfarmer-fearless, line 11: Everywhere you have `Type | null` you can replace with the shorthand version `Type?` with a question mark at the end.

tests/integration/playwright/client.ts
USZ7: Done.

tests/integration/playwright/test/upload-test.spec.ts
USZ7: Done.

tests/integration/playwright/package.json
USZ7: Playwright is used for development testing and automation, not in production, so it should be listed as a devDependency in package.json.

tests/integration/playwright/package.json
USZ7: Done.

tests/integration/playwright/package.json
USZ7: removed.

tests/integration/playwright/upload.ts
USZ7: Done.

tests/integration/playwright/upload.ts
USZ7: Done.

tests/integration/playwright/upload.ts
USZ7: Done.

tests/integration/playwright/upload.ts
USZ7: Done.

tests/integration/playwright/upload.ts
USZ7: Fixed.

tests/integration/playwright/login.ts
```diff
@@ -0,0 +1,14 @@
+import c, { LoginResponse } from './client';
+import config from './config';
+
+export async function loginAndGetToken() {
+  const [username, password, url] = config.validateEnv();
+
+  const loginResponse: LoginResponse | null = await c.login(username, password, url);
```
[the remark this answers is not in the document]
USZ7: tried to make the change as you suggested, but it caused an error.

tests/integration/playwright/config.ts
```diff
@@ -0,0 +1,53 @@
...
+dotenv.config({ path: "../../.env" });
+
+interface EnvConfig {
+  validateEnv: () => [string, string, string, string | undefined];
+}
+
+const config: EnvConfig = {
+  validateEnv: () => {
+    // Use CI/CD environment variables or fallback to .env variables
+    const username = process.env.SAMS_USERNAME || process.env.ACCOUNT_USERNAME;
```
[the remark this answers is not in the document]
USZ7: Done.

tests/integration/playwright/README.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+SAMS_USERNAME=""
+SAMS_PASSWORD=""
+UPLOAD_URL=""
+PS_API_URL=""
+```
+
+## Usage
+
+To run the script, invoke NodeJS passing the script name as its first argument, as in `node index.js`. You should get console output similar to this:
```
cfarmer-fearless, line 32: According to the docs, playwright tests are run with `npx playwright test`.  Can we run the tests with this command instead of running a javascript file directly?

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+      psApiUrl = `${PS_API_URL}/api/report/uploadId/${uploadId}`;
+
+    } catch (error) {
+      console.error("Setup failed:", error);
+      throw new Error("Failed to complete test setup.");
+    }
+  });
+
+  // Act & Assert 
+  test('upload and get trace response', async ({ request }) => {
```
cfarmer-fearless, line 50: The tests should cover the following cases:

- should return 500 with error message when destination ID not provided; PS API should return complete for metadata verify with equal error message
- should return 500 with error message when event type not provided; PS API should return complete for metadata verify with equal error message
- should return 500 with error message when destination ID and event type are mismatched; PS API should return complete for metadata verify with equal error message
- should return 500 when config schema definition invalid; PS API should return complete for metadata verify with equal error message

cfarmer-fearless (changes requested):

tests/integration/playwright/test/upload-test.spec.ts
USZ7: Done.

tests/integration/playwright/README.md
USZ7: Done.

tests/integration/playwright/client.ts
```diff
@@ -0,0 +1,39 @@
+import axios from "axios";
```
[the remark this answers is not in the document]
USZ7: Done.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  // Arrange
+  // Correctly use test.beforeAll at the describe level to perform setup actions
+  test.beforeAll(async ({ request }) => {
+    accessToken = await client.login(username, password);
+    if (!accessToken) throw new Error("Login failed.");
+  });
+
+  // Test case 1: Destination ID not provided
+  test('should return 500 when destination ID not provided', async ({ request }) => {
+    const metadata = createMetadata({ destinationId: undefined, eventType: "testevent1" });
```
cfarmer-fearless, line 62: This test should provide metadata without the `destinationId` field itself.  Here, you are providing the field with an undefined value.  Even though the value is undefined, you are still providing the field.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      await client.uploadFileAndGetId(accessToken, fileName, metadata);
+
+    } catch (error) {
+      assertErrorResponse(error, 500, "Not a recognized combination of meta_destination_id (undefined) and meta_ext_event (testevent1)");
+    }
+  });
+
+  // Test case 2: Event type not provided
+  test('should return 500 when event type not provided', async ({ request }) => {
+    const metadata = createMetadata({ destinationId: "dextesting", eventType: undefined });
```
cfarmer-fearless, line 74: Same here but for `eventType`.

tests/integration/playwright/test/upload-test.spec.ts
cfarmer-fearless: The assersions against the PS API response also need to verify the `issues` field contains the same error message that was given in the tus response.  It may be better to do these queries as part of your four tests above.

cfarmer-fearless (changes requested):

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,168 @@
+import { expect, test } from '@playwright/test';
+import dotenv from "dotenv";
```
cfarmer-fearless, line 2: Please use consistent quotation marks

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    test(name, async ({ request }) => {
+
+      try {
+        await client.uploadFileAndGetId(accessToken, fileName, metadata);
+
+      } catch (error) {
+        assertErrorResponse(error, expectedStatusCode, expectedErrorMessage);
+
+        const psApiUrl = `${PS_API_URL}/api/report/uploadId/${uploadId}`;
+        const response = await request.get(psApiUrl, { headers: { Authorization: `Bearer ${accessToken}` } });
```
cfarmer-fearless, line 121: The PS API does not need an auth token in the HTTP request.

cfarmer-fearless (changes requested):

tests/integration/playwright/test/upload-test.spec.ts
USZ7: Done.

tests/integration/playwright/test/upload-test.spec.ts
USZ7: Fixed.

tests/integration/playwright/test/upload-test.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    // Assert that the extracted status code matches the expected status code
+    expect(statusCode).toBe(expectedStatusCode);
+
+    // Further, assert that the error message contains the specific error detail
+    expect(errorMessage).toContain(expectedErrorMessageSubstring);
+
+    // Use RegExp.exec() to extract the uploadId
+    const uploadIdMatch = uploadIdRegex.exec(errorMessage);
+    uploadId = uploadIdMatch ? uploadIdMatch[1] : null;
```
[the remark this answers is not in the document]
USZ7: We can't set the uploadId from the result of client.uploadFileAndGetId if the upload fails. For non-happy paths, we must extract the uploadId from the error.

PR #247, opened 2024-02-29
Quickfix/ps api error handling

cfarmer-fearless:


upload-processor/BulkFileUploadFunctionApp/Model/DestinationAndEvents.cs
```diff
@@ -1,27 +1,25 @@
-using System.Text.Json.Serialization;
+using Newtonsoft.Json;
```
USZ7, line 1: Wouldn't this affect other bulk upload classes? We were using `System.Text.Json`. Just wanted to give you a heads-up.

USZ7 (approved):
lgtm!

upload-processor/BulkFileUploadFunctionApp/Model/DestinationAndEvents.cs
cfarmer-fearless: Good call!  I switched back to System.Text.Json because it's apparently faster.  I also removed newtonsoft from the project and converted everything over to System.Text.Json.

PR #250, opened 2024-03-01
UPLOAD-1027 Add global security stanza to the openapi.yml

whytheplatypus:
Before this the global or default security options for endpoints were unspecified in the documentation.

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -806,4 +806,5 @@ components:
             - type: number
 
 security:
-   - bearerAuth: []
+  - {}
+  - bearerAuth: []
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -805,4 +805,6 @@ components:
             - type: string
             - type: number
 
-
+security:
+  - {}
```
whytheplatypus, line 809: Based on [this documentation](https://swagger.io/specification/#optional-oauth2-security) this specifies optional security patterns used by each path. And can be specified more specifically per path.

AJKimpo (approved):
lgtm

PR #263, opened 2024-03-13
Upload 1034/file copy consistency

cfarmer-fearless:
This effort ultimately turned out to be a failure.  It was discovered that in order to copy a file across storage accounts using a scheduled lease (as opposed to streams), the storage account needs to be publicly accessible, or a SAS token with appropriate privileges is being used.  Neither of these approaches are options for the Upload API as they go against the CDC security policy for Azure storage.

I did, however, find some ways to clean up the code in a marginal amount and wanted to get that reviewed and merged anyway.

### DIFF
--- a/upload-processor/BulkFileUploadFunctionApp/Utils/BlobCopyHelper.cs
+++ b/upload-processor/BulkFileUploadFunctionApp/Utils/BlobCopyHelper.cs
@@ -20,11 +20,12 @@ namespace BulkFileUploadFunctionApp.Utils
             // Lease the source blob for the copy operation 
             // to prevent another client from modifying it.
             BlobLeaseClient lease = sourceBlob.GetBlobLeaseClient();
-            // Get the source blob's properties and display the lease state.
-            BlobProperties sourceProperties = await sourceBlob.GetPropertiesAsync();
-
+            
             try
             {
+                // Get the source blob's properties and display the lease state.
+                BlobProperties sourceProperties = await sourceBlob.GetPropertiesAsync();
+
                 _logger.LogInformation($"Checking if source blob with uri {sourceBlob.Uri} exists");
 
                 // Ensure that the source blob exists.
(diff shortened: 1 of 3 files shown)


upload-processor/BulkFileUploadFunctionApp/Utils/BlobCopyHelper.cs
```diff
@@ -15,8 +15,14 @@ public BlobCopyHelper(ILogger logger)
...
 
-        public async Task CopyBlobAsync(BlobClient sourceBlob, BlobClient destinationBlob, IDictionary<string, string> destinationMetadata, Uri? sourceSasBlobUri = null)
-        {
+        public async Task CopyBlobLeaseAsync(BlobClient sourceBlob, BlobClient destinationBlob, IDictionary<string, string> destinationMetadata, Uri? sourceSasBlobUri = null)
+        {
+            // Lease the source blob for the copy operation 
+            // to prevent another client from modifying it.
+            BlobLeaseClient lease = sourceBlob.GetBlobLeaseClient();
+            // Get the source blob's properties and display the lease state.
+            BlobProperties sourceProperties = await sourceBlob.GetPropertiesAsync();
```
USZ7, line 24: Is there any reason you removed them from the try-catch block? Also, wouldn't it be advisable to validate that sourceBlob, destinationBlob, and destinationMetadata are not null before we use them?

cfarmer-fearless: I want to use both of these variables in the `try` and `finally` block.  Good callout though because looks like `sourceBlob.GetPropertiesAsync()` can throw `RequestFailedException`.  So that should be moved back to inside the `try`, and the `sourceProperties` variable can be redefined within the `finally`.  Sound good?

cfarmer-fearless: `sourceBlob`, `destinationBlob`, and `destinationMetadata` are not defined as nullable in the function signature, so the function should assume they will never be null.

USZ7: Ok, agreed!

USZ7 (approved):
lgtm!

PR #271, opened 2024-03-26
upload-1101- Check invalid characters in filename metadata fields

USZ7:
- Check invalid characters in filename metadata fields


tus/file-hooks/metadata-verify/pre_create_bin.py
```diff
@@ -147,6 +147,10 @@ def report_verification_failure(messages, destination_id, event_type, meta_json)
 def stringify_error_messages(messages):
     return 'Found the following metadata validation errors: ' + ','.join(messages)
 
+def verify_filename(filename):
+    invalid_chars = set('<>:"/\\|?*')
```
cfarmer-fearless, line 151: Let's make this a global constant variable so it isn't instantiated every time.

tus/file-hooks/metadata-verify/pre_create_bin.py
```diff
@@ -147,6 +147,10 @@ def report_verification_failure(messages, destination_id, event_type, meta_json)
...
+    if any((c in invalid_chars) for c in filename):
+        raise ValueError(f"Filename contains invalid characters: {filename}")
```
cfarmer-fearless, line 153: Let's also return the set of invalid chars so the user knows what's not allowed.

cfarmer-fearless (changes requested):
Minor comments.  Looks good overall!  I like the use of a `set` so char lookups are fast :)

cfarmer-fearless (approved):
Looks great!

PR #277, opened 2024-04-01
upload-1114 implement feature flag for ps api in tus hooks

USZ7:
- Implement feature flagging for ps api in tus hooks

### DIFF
--- a/tus/file-hooks/post-create/post_create_bin.py
+++ b/tus/file-hooks/post-create/post_create_bin.py
@@ -9,12 +9,29 @@ from dotenv import load_dotenv
 
 from proc_stat_controller import ProcStatController
 
+from azure.appconfiguration import AzureAppConfigurationClient
+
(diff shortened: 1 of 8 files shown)


tus/file-hooks/post-create/post_create_bin.py
```diff
@@ -112,7 +129,9 @@ def main(argv):
 
     # Create upload trace.
     dest, event = get_required_metadata(metadata_json_dict)
-    post_create(dest, event, metadata_json_dict, tguid)
+
+    if processing_status_traces_enabled:
```
cfarmer-fearless, line 133: Might need to do a bit of refactoring here.  Looks like the reporting and tracing calls are too coupled in this hook.

cfarmer-fearless (commented):
Just some early feedback.  I like where this is going.  Don't forget to look at `post-receive` hook as well as it sends reports to PS API.

PR #282, opened 2024-04-04
UPLOAD-1181: add note that staging for `/status` endpoint is deprecated

dancing-goats:
https://cdc-dexops.atlassian.net/browse/UPLOAD-1181


docs/openapi.yml
whytheplatypus, line 1: Looks like [openapi added a `deprecated`](https://swagger.io/specification/#operationDeprecated) field. Worth adding for any tooling that consumes this.

whytheplatypus (changes requested):

docs/openapi.yml
dancing-goats: done, thanks for the 💡 idea!

whytheplatypus (approved):

PR #297, opened 2024-04-22
UPLOAD-1189 Adding a distributed lock to upload-server

whytheplatypus:
When an upload is in progress a lock needs to be obtained and managed across all running instances.


upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,157 @@
...
+	locker := &RedisLocker{
+		rs:    rs,
+		redis: client,
+	}
+	for _, option := range lockerOptions {
+		option(locker)
+	}
+	//defaults
+	if locker.logger == nil {
+		locker.logger = slog.Default()
```
whytheplatypus, line 37: Still need to finish the logger integration

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,157 @@
...
+	Listen(ctx context.Context, id string, callback func())
+	Request(ctx context.Context, id string)
+}
+
+type RedisLockExchange struct {
+	client *redis.Client
+}
+
+func (e *RedisLockExchange) channelName(id string) string {
+	return fmt.Sprintf("tusd_lock_release_request_%s", id)
```
whytheplatypus, line 53: Could make this prefix a module level `var` so that it's configurable if needed.

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,157 @@
...
+	go l.keepAlive(l.ctx)
+
+	return nil
+}
+
+func (l *redisLock) retryLock(ctx context.Context) error {
+	for {
+		l.exchange.Request(ctx, l.id)
+		select {
+		case <-time.After(500 * time.Millisecond):
```
whytheplatypus, line 122: similar to the prefix this should either be a pkg level `var` or something on the locker that has a default

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,157 @@
...
+			return handler.ErrLockTimeout
+		}
+	}
+}
+
+func (l *redisLock) keepAlive(ctx context.Context) {
+	//insures that an extend will be canceled if it's unlocked in the middle of an attempt
+	for {
+		select {
+		case <-time.After(time.Until(l.mutex.Until()) - 4*time.Second):
```
whytheplatypus, line 137: this timing is too arbitrary, needs to be built off the value of `Until`

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,157 @@
...
+			log.Println("lock was closed")
+			return
+		}
+	}
+}
+
+func (l *redisLock) Unlock() error {
+	_, err := l.mutex.Unlock()
+	if l.cancel == nil {
+		return errors.New("something's gone horribly wrong")
```
whytheplatypus, line 153: Need consistent error constants or types

upload-server/pkg/redislocker/redislocker_test.go
```diff
@@ -0,0 +1,163 @@
...
+		t.Error(err)
+	}
+	requestRelease := func() {
+		l.Unlock()
+	}
+	if err := l.Lock(ctx, requestRelease); err != nil {
+		t.Error(err)
+	}
+	t.Log("wait for refresh")
+	<-time.After(15 * time.Second)
```
whytheplatypus, line 93: need to configure this to meaningfully test but in a short amount of time

upload-server/cmd/cli/serve.go
```diff
@@ -33,7 +34,16 @@ func Serve(appConfig appconfig.AppConfig) (http.Handler, error) {
 	health.Register(storeHealthCheck)
 
 	// initialize locker
-	locker := memorylocker.New()
+	var locker handlertusd.Locker
+	locker = memorylocker.New()
```
dancing-goats, line 38: can consolidate the redundant assignment with:

```go
var locker handlertusd.Locker = memorylocker.New()
```

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+
+var (
+	PrefixString  = "tusd_lock_release_request_%s"
+	RetryInterval = 500 * time.Millisecond
+	LockExpiry    = 8 * time.Second
+)
+
+type LockerOption func(l *RedisLocker)
+
+func WithLogger(logger *slog.Logger) LockerOption {
```
dancing-goats, line 23: can `slog.Logger` be replaced with `log.Logger`?

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+			l.cancel()
+			if releaseRequested != nil {
+				releaseRequested()
+			}
+		}
+	}()
+	return nil
+}
+
+func (l *redisLock) lock(ctx context.Context) error {
```
dancing-goats, line 130: maybe rename `lock` function to `TryAcquireLock`?

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+type redisLock struct {
+	id       string
+	mutex    *redsync.Mutex
+	ctx      context.Context
+	cancel   func()
+	exchange LockExchange
+	logger   *slog.Logger
+}
+
+func (l *redisLock) Lock(ctx context.Context, releaseRequested func()) error {
```
dancing-goats, line 112: maybe rename `Lock` function to `AcquireAndManageLock`?

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+	"time"
+
+	"github.com/go-redsync/redsync/v4"
+	"github.com/go-redsync/redsync/v4/redis/goredis/v9"
+	"github.com/redis/go-redis/v9"
+	"github.com/tus/tusd/v2/pkg/handler"
+)
+
+var (
+	PrefixString  = "tusd_lock_release_request_%s"
```
dancing-goats, line 16: `LockReleaseRequestPrefix` instead of the less descriptive `PrefixString`?

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+	case <-c:
+		callback()
+	case <-ctx.Done():
+		return
+	}
+}
+
+func (e *RedisLockExchange) Request(ctx context.Context, id string) {
+
+	e.client.Publish(ctx, e.channelName(id), "please release")
```
dancing-goats, line 78: The message `release requested` is a bit more explicit about action needed.

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return locker, nil
+}
+
+type LockExchange interface {
+	Listen(ctx context.Context, id string, callback func())
+	Request(ctx context.Context, id string)
+}
+
+type RedisLockExchange struct {
```
dancing-goats, line 57: Is there a lock release notification? Would it be useful to detect when the lock is released by another process?

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (l *redisLock) retryLock(ctx context.Context) error {
+	for {
+		l.exchange.Request(ctx, l.id)
+		select {
+		case <-time.After(RetryInterval):
+			if err := l.lock(ctx); err != nil {
+				continue
+			}
+			return nil
```
dancing-goats, line 148: Should there be a case here to listen for lock release notifications?

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: not in this case, the structured nature of [log/slog](https://pkg.go.dev/log/slog) as well as it's extra functions e.g. `With` `Info` `Error` etc. are needed.

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: The name needs to conform to the [interface](https://pkg.go.dev/github.com/tus/tusd/v2@v2.4.0/pkg/handler#Lock) expected by tus

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: hmm maybe? the <name> -> error pair implies it could fail so we don't necessarily need the `try`.. `aquireLock` could be good, showing that it does have to request it from elsewhere

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: that's a really interesting idea, lets try that

upload-server/pkg/redislocker/redislocker.go
Acconut: This function's name is mandated by tusd's `Lock` interface: https://pkg.go.dev/github.com/tus/tusd/v2/pkg/handler#Lock

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,184 @@
...
+	}()
+	return nil
+}
+
+func (l *redisLock) aquireLock(ctx context.Context) error {
+	if err := l.mutex.TryLockContext(ctx); err != nil {
+		return err
+	}
+
+	l.ctx, l.cancel = context.WithCancel(context.Background())
```
Acconut, line 139:
```suggestion
	l.ctx, l.cancel = context.WithCancel(ctx)
```

Maybe it would make sense to bind `l.ctx` to `ctx`, so if `ctx` is cancelled `l.ctx` is cancelled as well, making sure that `l.exchange.Listen` and `l.keepAlive` are stopped.

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return nil
+}
+
+func (l *redisLock) retryLock(ctx context.Context) error {
+	for {
+		l.exchange.Request(ctx, l.id)
+		select {
+		case <-time.After(RetryInterval):
+			if err := l.aquireLock(ctx); err != nil {
```
Acconut, line 149: It would be good to distinguish between different error value here. An error indicating that another process holds the lock should yield in a retry, while other errors should be bubbled up to the main handler. The same would also apply to https://github.com/CDCgov/data-exchange-upload/pull/297/files#diff-783d3bbfaaa3b5128897203b93b326e6519991adf3a1070928b1d58b535b5969R117.

upload-server/pkg/redislocker/redislocker.go
Acconut: It would be possible to combine them together. If the process cannot acquire the lock because it is already held, it start polling and listening for release notifications. The polling ensures that we could obtain the lock if the holding process crashed, while the release notifications provides a faster response time because it does not have to wait for the next polling cycle. Which means that it allows you to have a higher polling interval while still maintaining a quick response for most uploads (assuming that server crashes are rare).

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,184 @@
...
+type RedisLockExchange struct {
+	client *redis.Client
+}
+
+func (e *RedisLockExchange) channelName(id string) string {
+	return fmt.Sprintf(LockExchangeChannelPrefix, id)
+}
+
+func (e *RedisLockExchange) Listen(ctx context.Context, id string, callback func()) {
+	psub := e.client.PSubscribe(ctx, e.channelName(id))
```
Acconut, line 70: This approach uses one channel per lock, so if tusd handles N uploads, it will subscribe to N channels. An alternative would be to have one channel for all uploads and the messages that are published to the channel identify the upload that should be stopped. If the resource usage of a subscription is non-negligible, this might be an alternative worth exploring. But I am not the Redis expert to judge this.

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: I assumed this context was centered around acquiring the lock, so it might timeout or be canceled too quickly? As opposed to a context that lives for the full lifecycle of an acquired lock.

upload-server/pkg/redislocker/redislocker.go
whytheplatypus: very good point.

upload-server/pkg/redislocker/redislocker.go
Acconut: I double checked and you are correct. `ctx`'s lifetime is for acquiring the lock. Please ignore my comment above.

upload-server/pkg/redislocker/redislocker.go
Acconut: You can call `callback` multiple times without issues. Every call after the first one will be a noop because the callback will just cancel a context and subsequent cancellations of a context are ignored in Go: https://github.com/tus/tusd/blob/7e24722afc727341298b8987c1c0a2347e9e994c/pkg/handler/unrouted_handler.go#L1346

PR #299, opened 2024-04-23
UPLOAD-1232 ehdi sender and producer id value updates

AJKimpo:
Updates to the ehdi-csv.json

- updated sender id values to state abbreviations and '-PHA' label
- updated data producer id values to state abbreviations and '-PHA' label


dancing-goats (approved):

USZ7 (approved):
lgtm!

PR #300, opened 2024-04-23
upload-1187 translate post finish to go

USZ7:
Translate 'post finish' to go


whytheplatypus (changes requested):
Great start please move this into the go code, similar to the one triggered here: https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/cmd/cli/hooks.go#L93

whytheplatypus:
This looks like most of the behavior was around traces, so do we still need this hook or can it be deprecated entirely?

USZ7:
> Great start please move this into the go code, similar to the one triggered here: https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/cmd/cli/hooks.go#L93

Done!

USZ7:
> This looks like most of the behavior was around traces, so do we still need this hook or can it be deprecated entirely?

@cosmycx  @cfarmer-fearless let's get your thoughts on this before we implement the code change.

cfarmer-fearless (changes requested):
This hook's only responsibility is to stop a span in PS API, which is deprecated.  Therefore, this hook can be removed entirely and does not need to be translated to GoLang.

USZ7:
> This hook's only responsibility is to stop a span in PS API, which is deprecated. Therefore, this hook can be removed entirely and does not need to be translated to GoLang. This PR can be closed, and corresponding Jira ticket canceled.

@cfarmer-fearless Gotcha, thanks! I'll close it out. CC @whytheplatypus

USZ7:
The 'post finish' hook can be removed entirely and does not need to be translated to GoLang. Based on the feedback, this PR can be closed.

PR #301, opened 2024-04-23
UPLOAD-1221 Upload Info Endpoint

whytheplatypus:


upload-server/cmd/cli/info.go
```diff
@@ -0,0 +1,195 @@
...
+	id := r.PathValue("UploadID")
+
+	fileInfo, err := ih.inspecter.InspectInfoFile(r.Context(), id)
+	if err != nil {
+		http.Error(rw, err.Error(), getStatusFromError(err))
+		return
+	}
+	uploadedFileInfo, err := ih.inspecter.InspectUploadedFile(r.Context(), id)
+	if err != nil {
+		http.Error(rw, err.Error(), getStatusFromError(err))
```
whytheplatypus, line 151: Include partial information in the body of this one. Instead of the error maybe an InfoResponse with just the Manifest?

upload-server/cmd/cli/info.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return uploadedFileInfo, nil
+}
+
+func (ih *InfoHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	id := r.PathValue("UploadID")
+
+	fileInfo, err := ih.inspecter.InspectInfoFile(r.Context(), id)
+	if err != nil {
+		http.Error(rw, err.Error(), getStatusFromError(err))
```
whytheplatypus, line 146: Should probably make the response body more generic here, just 404 not found or something like that.

upload-server/cmd/cli/info.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, errors.Join(err, ErrNotFound)
+	}
+	uploadedFileInfo := map[string]any{
+		"updated_at": fi.ModTime(),
+		"size_bytes": fi.Size(),
+	}
+	return uploadedFileInfo, nil
+}
+
+func (aui *AzureUploadInspector) InspectInfoFile(c context.Context, id string) (map[string]any, error) {
```
whytheplatypus, line 97: Now that this is established extract it into it's own module. For example `azureinspector` and `fileinspector`. A similar pattern is used by tusd [here](https://github.com/tus/tusd/tree/main/pkg).

upload-server/cmd/cli/info.go
cfarmer-fearless: Done

upload-server/cmd/cli/info.go
cfarmer-fearless: I've created these two modules as suggested.  I also moved all common public types and variables to `pkg/info.go`.  Lmk if there's a better approach.

cfarmer-fearless (approved):

PR #302, opened 2024-04-23
UPLOAD-1186 Translate post-receive hook to go

dtisza1:
Translate the [post-receive hook](https://github.com/CDCgov/data-exchange-upload/blob/main/tus/file-hooks/upload-status/post_receive_bin.py) that submits data to PS API into go and include it in the Upload Go binary.

CURRENT STATUS:

- Added and registered postReceive() as a tusd plugin type hook function.

-  This is testing OK with the server running via:
```
go run ./cmd/main.go -appconf=./configs/local/local.env
```
- Remove of HookConfigLoader, since the config isn't used for this hook.
- Created initial shell script code in Go.
- NEXT:  Will be adding migrated python code logic next.

### DIFF
--- a/upload-server/cmd/cli/hooks.go
+++ b/upload-server/cmd/cli/hooks.go
@@ -89,14 +89,14 @@ func postReceiveHook(event handler.HookEvent) (hooks.HookResponse, error) {
 	uploadId := event.Upload.ID
 	uploadSize := event.Upload.Size
 	uploadOffset := event.Upload.Offset
-	manifest := event.Upload.MetaData
+	uploadMetadata := event.Upload.MetaData
 
 	logger.Info(
 		"[PostReceive]: event.Upload values",
-		" manifest: ", manifest,
-		" uploadId: ", uploadId,
-		" uploadSize: ", uploadSize,
-		" uploadOffset: ", uploadOffset,
+		"uploadMetadata", uploadMetadata,
+		"uploadId", uploadId,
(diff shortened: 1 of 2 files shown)


upload-server/cmd/cli/hooks.go
```diff
@@ -70,6 +72,40 @@ func (l *AzureConfigLoader) LoadConfig(ctx context.Context, path string) ([]byte
 	return io.ReadAll(downloadResponse.Body)
 }
 
+// TODO: Combine somewhere for all hooks needing ConfigLoader
+//
+//	       like preCreate?
+//		      Is currently a duplicate for  metadata.SenderManifestVerification
+type HookConfigLoader struct {
+	Loader validation.ConfigLoader
```
whytheplatypus, line 80: It doesn't seem like post-receive needs to load the configs, so you can probably drop this to simplify things for now

whytheplatypus (commented):
solid start, you've got the hook registered and a general structure laid out well

upload-server/cmd/cli/hooks.go
tedtally: @whytheplatypus  Since we don't need to load the configs, is the main purpose for post-receive hook is to receive the file , get get meta data and other info, generate the report , and send a report to the service bus?

dtisza1: @whytheplatypus Just removed that struct. TY!

PR #315, opened 2024-04-29
UPLOAD-1258/global-timestamp

cfarmer-fearless:
This PR adds a global timestamp field to the sender manifest, which ultimately gets attached to the metadata of the file.  This global timestamp is generated when an upload is being initialized.  It will be used as a global time anchor for searching for uploads across a time range.  It is stored in ISO format in UTC.

### DIFF
--- a/upload-server/internal/metadata/metadata.go
+++ b/upload-server/internal/metadata/metadata.go
@@ -134,15 +134,13 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent) (hooks.Hook
 	return resp, nil
 }
 
-// type HandlerFunc func(event handler.HookEvent) (hooks.HookResponse, error)
-
 func WithTimestamp(next prebuilthooks.HookHandlerFunc) prebuilthooks.HookHandlerFunc {
 	return func(event handler.HookEvent) (hooks.HookResponse, error) {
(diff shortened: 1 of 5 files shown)


upload-server/internal/metadata/metadata.go
```diff
@@ -131,3 +133,21 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent) (hooks.Hook
...
+
+// type HandlerFunc func(event handler.HookEvent) (hooks.HookResponse, error)
+
+func WithTimestamp(next prebuilthooks.HookHandlerFunc) prebuilthooks.HookHandlerFunc {
+	return func(event handler.HookEvent) (hooks.HookResponse, error) {
+		timestamp := time.Now().Format(time.RFC3339)
+		logger.Info("adding global timestamp", "timestamp", timestamp)
+
+		manifest := event.Upload.MetaData
+		manifest["global_timestamp"] = timestamp
```
whytheplatypus, line 145: sounds like for the key people are thinking `dex_ingest_datetime`

cfarmer-fearless: Sounds good

whytheplatypus (approved):

PR #319, opened 2024-05-01
Handle both trailing slash and not for uploads

whytheplatypus:
Before this a client that didn't handle 301 redirects well would fail if they used a path like `/files` this change makes it work easily for both.


cfarmer-fearless (changes requested):
Can we add a small test for this?  All the tests are targeting `/files/`.  Seems like we can create a small happy path test that targets `/files` too.

cfarmer-fearless (approved):

PR #340, opened 2024-05-07
clean up ps health check

cosmycx:


whytheplatypus (approved):
Something odd seems to be going on with github, and how it's displaying the diffs / commit history...but other than that this looks good to me

PR #353, opened 2024-05-15
lowercase config filename

cfarmer-fearless:
This patch is to address the case sensitivity of upload config files.  Before this patch, destination/datastream and event/route values in sender manifests needed to be an exact match as these values are used to read json config files, which is case sensitive.  After this patch, the values users put in their sender manifest for these fields are no longer case sensitive.  This is because all json config files have been lowercased, and the values used to get those config files will be lowercased in code.

By this, we can create consistent casing within the Upload API for upload configs, without breaking users who are sending manifests with capital letters.

Main motivation here is we did this for pulsenet, but now we need to do it for the rest of the configs.  In other words, all configs need to follow same naming convention.


AJKimpo (approved):
approved

PR #360, opened 2024-05-17
UPLOAD-1341 Support backwards compatible upload config filenames

cfarmer-fearless:
This patch is to address a potential issue when transitioning senders from v1 to v2 sender manifest definitions.  Currently, we have a limitation where the v1 and v2 config filenames must match.  If they don't, the upload processor will throw an error as it will be unable to find the v2 config file given v1 use case identifiers if the identifiers are different across versions.

Ex:
v1 destination ID is `myprogram` and event type is `superawesomeevent`.  
v2 data stream ID is `mynewprogram` and data stream route is `superroute`.

Given a v1 manifest, the upload processor will look for upload config files named `myprogram-superawesomeevent.json` for both v1 and v2 configs.  This will result in an error.

This patch adds a field to the upload config called `compat_config_filename` that links a v1 config file to a v2 config file.  This is an optional field.  If it is provided, the upload processor will use it to find the v2 config file.  If it's not, it will revert to the old behavior and use the use case identifiers.

A good test case for this is the AIMS CELR use case.  In v1, their config filename is aims-celr-csv.json.  In v2, it's just celr-csv.json.  Uploading a file for aims-celr-csv should map to celr-csv.

### DIFF
--- a/upload-configs/v1/aims-celr-csv.json
+++ b/upload-configs/v1/aims-celr-csv.json
@@ -73,5 +73,6 @@
 		"targets": [
 			"routing"
 		]
-	}
+	},
+	"compat_config_filename": "celr-csv.json"
 }
--- a/upload-configs/v1/aims-celr-hl7.json
+++ b/upload-configs/v1/aims-celr-hl7.json
@@ -73,5 +73,6 @@
 		"targets": [
 			"routing"
 		]
-	}
+	},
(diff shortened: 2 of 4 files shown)


USZ7 (approved):

tedtally (commented):
lgtm

PR #361, opened 2024-05-18
Use 2 channels per lock for exchange

whytheplatypus:
Also sends fewer release requests

### DIFF
--- a/upload-server/pkg/redislocker/redislocker.go
+++ b/upload-server/pkg/redislocker/redislocker.go
@@ -3,6 +3,7 @@ package redislocker
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"time"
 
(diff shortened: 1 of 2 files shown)


upload-server/pkg/redislocker/redislocker.go
```diff
@@ -171,30 +167,33 @@ func (l *redisLock) aquireLock(ctx context.Context) error {
...
-			return errs
-		}
-		select {
-		case <-c:
-			continue
-		case <-time.After(RetryInterval):
-			continue
-		case <-ctx.Done():
-			return errors.Join(errs, handler.ErrLockTimeout)
+		if !errors.Is(err, handler.ErrFileLocked) {
```
whytheplatypus, line 190: at this point we can just exit with the error.

upload-server/pkg/redislocker/redislocker.go
```diff
@@ -221,8 +220,12 @@ func (l *redisLock) Unlock() error {
 	if l.cancel != nil {
 		defer l.cancel()
 	}
-	_, err := l.mutex.Unlock()
-	if e := l.exchange.Release(l.ctx, l.id); e != nil {
+	b, err := l.mutex.UnlockContext(l.ctx)
+	l.logger.Info("lock released", "err", err, "released", b)
```
whytheplatypus, line 224: Need to handle the case where b is false and the lock isn't released. Look into tus code to see what happens if "unlock" errors

whytheplatypus: nothing is done with errors from Unlock

cfarmer-fearless (approved):

PR #370, opened 2024-05-22
Upload 1261 review audit logs v2 server

tedtally:


upload-server/internal/handlertusd/handlertusd.go
```diff
@@ -20,6 +20,8 @@ func init() {
 	type Empty struct{}
 	pkgParts := strings.Split(reflect.TypeOf(Empty{}).PkgPath(), "/")
 	// add package name to app logger
+	// TODO: replace logger with slog.logger?
+	// Why are we using slogerxexp instead of sloger?
```
whytheplatypus, line 24: tus requires that specific type.

upload-server/internal/handlertusd/handlertusd.go
```diff
@@ -20,6 +20,8 @@ func init() {
 	type Empty struct{}
 	pkgParts := strings.Split(reflect.TypeOf(Empty{}).PkgPath(), "/")
```
whytheplatypus, line 21: look into replace these calls in the `init`, we maybe able to get the same or better debug information by enabling `AddSource` at the top level https://pkg.go.dev/log/slog#HandlerOptions

upload-server/internal/metadata/metadata.go
```diff
@@ -219,10 +219,9 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 }
 
 func (v *HookEventHandler) WithUploadID(event handler.HookEvent, resp hooks.HookResponse) (hooks.HookResponse, error) {
-
 	tuid := Uid()
 	resp.ChangeFileInfo.ID = tuid
-
+	logger := WithTGUID(tuid)
```
whytheplatypus, line 224: this is hiding that it's using the package level logger to create this. Consider exposing that for clarity: `l := logger.With("tguid", tguid)`

whytheplatypus (changes requested):
solid draft. Can you double check that the logs comming from the tus lib are using the logger we pass to it and are being output as json? I think last I looked they weren't coming through as json objects.

upload-server/internal/metadata/metadata.go
whytheplatypus: can also make the key a constant.. so something like `logger.With(reports.TGUID_KEY, tuid)`

upload-server/internal/handlertusd/handlertusd.go
tedtally: ok.

upload-server/internal/metadata/metadata.go
tedtally: So would this replace the Report's existing uploadID property with TGUID_KEY?

upload-server/internal/handlertusd/handlertusd.go
tedtally: removed the comments

upload-server/internal/metadata/metadata.go
tedtally: Added the constant to the models.go file then updated the calls with logger := logger.With(models.TGUID_KEY, tuid).

upload-server/internal/metadata/metadata.go
```diff
@@ -172,7 +172,7 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 	if tuid == "" {
 		return resp, errors.New("no Upload ID defined")
 	}
-
+	logger := logger.With(models.TGUID_KEY, tuid)
```
cfarmer-fearless, line 176: Do you need to call `logger.With` multiple times?  Can you just call it as soon as you have the tguid and it'll be set for the lifetime of the upload?

upload-server/internal/models/consts.go
```diff
@@ -23,4 +23,5 @@ const (
 	TARGET_DEX_EDAV   = "dex_edav"
 
 	EVENT_UPLOAD_ID = "event.Upload.ID"
+	TGUID_KEY       = "tguid"
```
dtisza1, line 26: Should we name this key "id" to match the naming in the tusd lib related logging? This would provide grouping using just one key name.

dtisza1 (changes requested):
Naming change suggestion.

upload-server/internal/metadata/metadata.go
tedtally: That struck me as well, and at first I had a func that was called after init that took the responsibility of calling Uid() away from WithUploadId but the func was hiding that it was using the package level logger to create the tguid append (which we were informed was bad form). We'll rework this to make this more efficient

upload-server/internal/models/consts.go
tedtally: I think ID may be too ambiguous, in the event we start to incorporate other unique identifiers this may make things more complicated

upload-server/internal/metadata/metadata.go
```diff
@@ -219,10 +218,13 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 }
 
 func (v *HookEventHandler) WithUploadID(event handler.HookEvent, resp hooks.HookResponse) (hooks.HookResponse, error) {
-
 	tuid := Uid()
 	resp.ChangeFileInfo.ID = tuid
 
+	if sloger.DefaultLogger != nil {
+		logger = sloger.DefaultLogger.With(models.TGUID_KEY, tuid, "TESTING", "123")
```
cfarmer-fearless, line 225: Is the "TESTING" "123" necessary?

dtisza1: Oops. Forgot to remove that :-)

upload-server/cmd/cli/logger.go
```diff
@@ -40,6 +40,9 @@ func AppLogger(appConfig appconfig.AppConfig) *slog.Logger {
 
 	appLogger := logger.With(
 		slog.Group("app_info",
+			slog.String("System", "DEX"),
```
cfarmer-fearless, line 43: Something must be up here.  Getting logs with a weird structure:
```
{"time":"2024-05-24T11:37:25.581131-04:00","level":"INFO","msg":"ChunkWriteStart","!BADKEY":{"Key":"app_info","Value":{}},"buildInfo.Main.Path":"","pkg":"main","method":"PATCH","path":"b2be5ea125156135566f26dabbbf99f8","requestId":"","id":"b2be5ea125156135566f26dabbbf99f8","maxSize":100000,"offset":900000}
```
Notice the `!BADKEY` key name.

cfarmer-fearless (changes requested):

upload-server/internal/handlertusd/handlertusd.go
tedtally: done. Included AddSource to HandlerOptions in Logger.go

upload-server/internal/metadata/metadata.go
dtisza1: @cfarmer-fearless We integrated the logger.With() with the UploadWithID() hook handler only. The key values persist in the logging after that.

PR #375, opened 2024-05-27
Upload 1361/smoke test use case inconsistency

cfarmer-fearless:
Refactor of the Upload API smoke tests.  Uses TestNG data providers for parametric test data instead of the `@Parameter` decorator pattern.  Goal is to make it easier to add new test data for new use cases and scenarios, especially for sender manifests.  Instead of using the directory tree to categories sender manifests by use case, we instead have buckets of manifests that mix use cases, and group manifests by the test scenario or capability.  For example, we'll have a single JSON file with a list of all valid V1 sender manifests.  

This refactor also removes the need for the `-Penv` command line arg.  The `-PuseCases` command line arg is unchanged at a usage level.  However, it is handled deeper at a data provider level instead of a gradle config level.


tests/smoke/kotlin/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ plugins {
 }
 
 kotlin {
-    jvmToolchain(17)
+    jvmToolchain(11)
```
cfarmer-fearless: Please revert this back to 17

tests/smoke/kotlin/build.gradle.kts
```diff
@@ -60,9 +60,8 @@ tasks.test {
 //    }
 }
 
-
 tasks.withType<KotlinCompile> {
-    kotlinOptions.jvmTarget = "17"
+    kotlinOptions.jvmTarget = "11"
```
cfarmer-fearless, line 64: Please revert this back to 17

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
```diff
@@ -1,151 +1,78 @@
...
-    fun beforeMethod() {
+    fun setupUpload(context: ITestContext) {
+        testContext = context
         uploadClient = UploadClient(EnvConfig.UPLOAD_URL, authToken)
     }
 
     @Test(
         groups = [Constants.Groups.METADATA_VERIFY],
-        dataProvider = "versionProvider",
+        dataProvider = "validManifestV1Provider",
```
cfarmer-fearless, line 43: This can use the `validManifestAllProvider`

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
```diff
@@ -1,151 +1,78 @@
...
-        expectedExceptionsMessageRegExp = "unexpected status code \\(400\\).*",
-        dataProvider = "versionProvider",
+        dataProvider = "validManifestAllProvider",
         dataProviderClass = DataProvider::class
     )
-    fun shouldReturnErrorWhenFilenameContainsInvalidChars(version: String) {
-        metadata = Metadata.getSenderManifest(version, useCase, senderManifestInvalidFilename)
-        uploadClient.uploadFile(testFile, metadata)
-    }
+    fun shouldValidateMetadataWithManifest(manifest: HashMap<String, String>) {
```
cfarmer-fearless, line 58: Perhaps this test can be combined with the one above

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
```diff
@@ -1,151 +1,78 @@
...
     }
 
     @Test(
         groups = [Constants.Groups.METADATA_VERIFY],
-        expectedExceptions = [ProtocolException::class],
-        expectedExceptionsMessageRegExp = "unexpected status code \\(400\\).*",
-        dataProvider = "versionProvider",
-        dataProviderClass = DataProvider::class
-    )
-    fun shouldReturnErrorWhenDestinationIDNotProvided(version: String) {
```
cfarmer-fearless, line 84: We can probably combine all of these tests into one that checks for required fields, and then have a data provider that gives a variety of manifests missing required fields, such as destination ID, event, filename, data stream ID, data stream route, etc.

tests/smoke/kotlin/src/test/kotlin/util/DataProvider.kt
```diff
@@ -44,7 +44,33 @@ class DataProvider {
             return manifests.map { arrayOf(it) }.toTypedArray()
         }
 
-        private fun filterByUseCases(useCases: List<String>, manifests: List<HashMap<String, String>>): List<HashMap<String, String>> {
+        @DataProvider(name = "validManifestAllProviderWithVersion")
+        @JvmStatic
+        fun validManifestAllProviderWithVersion(): Array<Array<Any>> {
```
cfarmer-fearless: Not sure if we need this.  We should be able to infer the version from the manifest itself.

tests/smoke/kotlin/src/test/kotlin/util/Metadata.kt
```diff
@@ -71,5 +73,27 @@ class Metadata {
         private fun getMetadataPath(version: String, useCase: String, manifest: String): String {
             return "properties/$version/$useCase/$manifest"
         }
+
+        fun readMetadataFromJsonFile(useCase: String): HashMap<String, String> {
```
cfarmer-fearless, line 77: What's the purpose of this function?  Looks like it searches all of the manifest test json data and returns a map that is a combination of all of the k/v pairs of manifests that match the given use case.  What is that used for?

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
GSA2024: ok, updated to all providers

tests/smoke/kotlin/build.gradle.kts
GSA2024: reverted back to version17

tests/smoke/kotlin/build.gradle.kts
GSA2024: reverted

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
GSA2024: As discussed, we plan to put all negative tests with `invalid_manifests<version>.json` files, right?

tests/smoke/kotlin/src/test/kotlin/util/DataProvider.kt
GSA2024: This is already removed with latest commit

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
GSA2024: Done

tests/smoke/kotlin/src/test/kotlin/MetadataVerify.kt
cfarmer-fearless: Yes.  Not sure yet if the negative tests need to be broken up by version though.  Let's see

PR #377, opened 2024-05-29
Go Upload Post Processing

whytheplatypus:
Goals:
- [ ] be able to recover at any given point
   - [x] preperation of file for dex (getting metadata into azure)
   - [x] delivery of files to each target
   - [ ] trigger recovery
   - [ ] automatic recover on restart (turning it off and back on again should do something)
- [x] strong observability into what has happened

### DIFF
--- a/upload-server/cmd/cli/hooks.go
+++ b/upload-server/cmd/cli/hooks.go
@@ -10,6 +10,7 @@ import (
 	azureloader "github.com/cdcgov/data-exchange-upload/upload-server/internal/loaders/azure"
 	fileloader "github.com/cdcgov/data-exchange-upload/upload-server/internal/loaders/file"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/metadata"
+	"github.com/cdcgov/data-exchange-upload/upload-server/internal/postprocessing"
 	azurereporters "github.com/cdcgov/data-exchange-upload/upload-server/internal/reporters/azure"
 	filereporters "github.com/cdcgov/data-exchange-upload/upload-server/internal/reporters/file"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/storeaz"
@@ -95,7 +96,7 @@ func PrebuiltHooks(appConfig appconfig.AppConfig) (tusHooks.HookHandler, error) 
 	handler.Register(tusHooks.HookPostCreate, hookHandler.PostCreate)
 	// note that tus sends this to a potentially blocking channel.
 	// however it immediately pulls from that channel in to a goroutine..so we're good
-	handler.Register(tusHooks.HookPostFinish, manifestValidator.Hydrate, copier.Merge, copier.Route)
+	handler.Register(tusHooks.HookPostFinish, manifestValidator.Hydrate, postprocessing.Merge, postprocessing.Deliver)
 
 	return handler, nil
 }
--- a/upload-server/cmd/cli/hooks.go
+++ b/upload-server/cmd/cli/hooks.go
@@ -2,10 +2,7 @@ package cli
 
 import (
 	"context"
-	"errors"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"net"
-	"net/http"
 	"os"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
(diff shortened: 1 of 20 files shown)


upload-server/cmd/cli/hooks.go
```diff
@@ -50,12 +54,27 @@ func PrebuiltHooks(appConfig appconfig.AppConfig) (tusHooks.HookHandler, error)
...
+	postprocessing.RegisterTarget("edav", &postprocessing.FileDeliverer{
+		ToPath: appConfig.LocalEDAVFolder,
+		From:   os.DirFS(appConfig.LocalFolderUploadsTus + "/" + appConfig.TusUploadPrefix),
+	})
+
+	postprocessing.RegisterTarget("routing", &postprocessing.FileDeliverer{
+		ToPath: appConfig.LocalROUTINGFolder,
+		From:   os.DirFS(appConfig.LocalFolderUploadsTus + "/" + appConfig.TusUploadPrefix),
+	})
+
```
whytheplatypus, line 71: this hooks file is starting to get big, might be the right time to start separating concerns across some files in the `cli` directory. May be able to do this now that file hooks are completely deprecated.

upload-server/cmd/cli/hooks.go
```diff
@@ -86,14 +105,45 @@ func PrebuiltHooks(appConfig appconfig.AppConfig) (tusHooks.HookHandler, error)
...
+					if storageErr.StatusCode == http.StatusNotFound {
+						logger.Info("creating dex-checkpoint container")
+						_, err := dexCheckpointContainerClient.Create(context.TODO(), nil)
+						if err != nil {
+							logger.Error("failed to create dex checkpoint container")
+							return nil, err
+						}
+					}
+				}
+			}
```
whytheplatypus, line 131: wonder what would be needed to allow the possible delivery points to be dynamic... or if that even makes any kind of sense as a feature.

upload-server/internal/postprocessing/deliver.go
```diff
@@ -0,0 +1,143 @@
...
+func (ad *AzureDeliverer) Deliver(tuid string, manifest map[string]string) error {
+	// Get blob src blob client.
+	// TODO get tus prefix from app config.  Place in azure deliverer.
+	srcBlobClient := ad.FromContainerClient.NewBlobClient(ad.TusPrefix + "/" + tuid)
+	// Get filename from metadata.
+	filename := metadata.GetFilename(manifest)
+	tokens := strings.Split(filename, ".")
+	extension := ""
+	if len(tokens) > 1 {
+		extension = "." + tokens[len(tokens)-1]
```
whytheplatypus, line 98: can use [`filepath.Ext`](https://pkg.go.dev/path/filepath@go1.22.3#Ext)

upload-server/internal/postprocessing/deliver.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// Load config from metadata.
+	// TODO create a utility function from this.
+	path, err := metadata.GetConfigIdentifierByVersion(context.TODO(), manifest)
```
whytheplatypus, line 103: move the context.TODO() to a single call at the top of the function and reuse the result. This will make the move to a real context easier.

upload-server/internal/postprocessing/deliver.go
```diff
@@ -0,0 +1,143 @@
...
+		manifestPointer[k] = &value
+	}
+	logger.Info("starting copy from", "src", srcBlobClient.URL(), "to dest", destBlobClient.URL())
+	resp, err := destBlobClient.StartCopyFromURL(context.TODO(), srcBlobClient.URL(), &blob.StartCopyFromURLOptions{Metadata: manifestPointer})
+	if err != nil {
+		return err
+	}
+
+	status := *resp.CopyStatus
+	for status == "pending" {
```
whytheplatypus, line 131: looks like `blob.CopyStatusTypePending` can be used as a constant for this check

upload-server/internal/metadata/metadata.go
```diff
@@ -217,7 +219,138 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
...
+	TusPrefix       string
+}
+
+type Appender interface {
+	Append(event handler.HookEvent, resp hooks.HookResponse) (hooks.HookResponse, error)
+}
+
+type MetadataAppender struct {
+	Appender Appender
+}
```
whytheplatypus, line 302: This can either be removed since it's a wrapper around an interface without any methods, or an Append method can be added that calls the "internal" append function.

If it's removed then when it's used in hooks.go a variable can be create of type `Appender`

upload-server/internal/metadata/metadata.go
```diff
@@ -217,7 +219,138 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
...
+
+func (fa *FileMetadataAppender) Append(event handler.HookEvent, resp hooks.HookResponse) (hooks.HookResponse, error) {
+	tuid := event.Upload.ID
+	if resp.ChangeFileInfo.ID != "" {
+		tuid = resp.ChangeFileInfo.ID
+	}
+	if tuid == "" {
+		return resp, errors.New("no Upload ID defined")
```
whytheplatypus, line 310: make this error a top level value. `ErrNoUploadID` or something

upload-server/internal/postprocessing/deliver.go
```diff
@@ -0,0 +1,184 @@
...
+	targets[name] = d
+}
+
+type Deliverer interface {
+	Deliver(tuid string, metadata map[string]string) error
+}
+
+// target may end up being a type
+func Deliver(tuid string, manifest map[string]string, target string) error {
+	ctx := context.TODO()
```
whytheplatypus, line 44: Even though there isn't a good top level context to pass around yet start pulling these TODO's up into one. So this and the other `Deliver` function should take a context as their first arg.

upload-server/internal/postprocessing/deliver.go
```diff
@@ -0,0 +1,184 @@
...
+
+	switch target {
+	case "edav":
+	case "routing":
+		if config.Copy.FolderStructure == "date_YYYY_MM_DD" {
+			// Get UTC year, month, and day
+			t := time.Now().UTC()
+			prefix = fmt.Sprintf("%d/%02d/%02d/", t.Year(), t.Month(), t.Day())
+		}
+	}
```
whytheplatypus, line 181: extract this and have it return the prefix. Should be able to add some tests then, this is a good one for a unit test since it's a little finicky

cfarmer-fearless: done

upload-server/internal/metadata/metadata.go
cfarmer-fearless: Removed

upload-server/internal/metadata/metadata.go
cfarmer-fearless: done

upload-server/internal/postprocessing/deliver.go
cfarmer-fearless: done

cfarmer-fearless (approved):

PR #381, opened 2024-06-03
Use workers for post processing

whytheplatypus:
Not to be considered until after #377


cfarmer-fearless:
Closing in favor of #396

PR #385, opened 2024-06-08
ED3N sender manifest config update

AJKimpo:
## Updates

- updated ED3N config: custom metadata field names, allowed values addition for custom metadata field, allowed value updates for sender id, data producer id, and jurisdiction


cfarmer-fearless (commented):
Would any of this information be considered PII?

AJKimpo:
patient dob metadata field removed

cfarmer-fearless (approved):

PR #387, opened 2024-06-11
fix check for service principle

cfarmer-fearless:
Check env vars from OS for service principle creds instead of app conf object.


dtisza1 (approved):
This makes sense and looks good to me!

PR #392, opened 2024-06-12
Add Health check dependencies on edav and routing storage

USZ7:
- Add health check dependencies on edav and routing.


cfarmer-fearless:
Closing in favor of #393

PR #393, opened 2024-06-14
Delivery Connection Resiliance and Health Check

cfarmer-fearless:
Idea here is to relate registration of file deliverers and registration of health checks on the respected delivery storage targets.  Eg: Registering an EDAV deliverer should also register a health check on the EDAV storage (file or azure).

### DIFF
--- a/upload-server/cmd/cli/hooks.go
+++ b/upload-server/cmd/cli/hooks.go
@@ -1,7 +1,6 @@
 package cli
 
 import (
-	"context"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/upload"
 	"os"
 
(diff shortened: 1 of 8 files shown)


cfarmer-fearless:
Currently, not sure if deliverer and health check registration should be at the serve level or hook level.

upload-server/cmd/cli/hooks.go
```diff
@@ -43,20 +44,18 @@ func PrebuiltHooks(appConfig appconfig.AppConfig) (tusHooks.HookHandler, error)
...
+		return nil, err
+	}
+	edavFileDeliverer, err := postprocessing.NewFileDeliverer(ctx, "edav", &appConfig)
+	if err != nil {
+		return nil, err
+	}
+	routingFileDeliverer, err := postprocessing.NewFileDeliverer(ctx, "routing", &appConfig)
+	if err != nil {
+		return nil, err
+	}
```
dtisza1, line 58: @cfarmer-fearless  I think these lines should be moved inside the else block on line 105, since they are not used when `appConfig.AzureConnection != nil`. Also the `metadataAppender` related lines above that.

dtisza1 (changes requested):
@cfarmer-fearless Just had a few changes/additions, if they make sense.

For local file system storage, the following empty files could be added so that the related directories exist:

uploads/dex/.gitkeep
uploads/edav/.gitkeep
uploads/routing/.gitkeep

I was able to test locally and the code looks good to me!

upload-server/cmd/cli/hooks.go
cfarmer-fearless: Done

dtisza1 (approved):
@cfarmer-fearless Those changes and the PR look great!

Is working as expected locally too.

PR #396, opened 2024-06-18
Upload 1401/post processing workers

cfarmer-fearless:
Copy files to there targets via workers running on separate threads.  Wait for all workers to complete in event of shutdown to avoid partial copies.


upload-server/cmd/cli/processor.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func StartProcessorWorkers(ctx context.Context) (chan postprocessing.Event, *sync.WaitGroup) {
+	var wg sync.WaitGroup
+	numWorkers := runtime.NumCPU()
+	c := make(chan postprocessing.Event, numWorkers)
+	for i := 0; i < numWorkers; i++ {
+		go postprocessing.Worker(ctx, c, &wg)
+	}
+	return c, &wg
```
whytheplatypus, line 18: Let this function be blocking with the wg rather than returning it. (this involves adding a wg.Add(numWorkers) line and then a wrapper around the `postprocessing.Worker` call that calls wg.Done().

upload-server/cmd/cli/serve.go
```diff
@@ -67,17 +68,18 @@ func Serve(appConfig appconfig.AppConfig) (http.Handler, error) {
 	err = InitReporters(appConfig)
 
 	// get and initialize tusd hook handlers
-	hookHandler, err := GetHookHandler(context.TODO(), appConfig)
+	postProcessingChannel, postProcessWaitGroup := StartProcessorWorkers(ctx)
```
whytheplatypus, line 71: bring this up into main. This is because this is conceptually a fully separate service than the web server.

upload-server/cmd/main.go
```diff
@@ -120,6 +120,11 @@ func main() {
 	signal.Notify(sigint, os.Interrupt)
 	<-sigint
 
+	// ------------------------------------------------------------------
+	// 	Wait for workers
+	// ------------------------------------------------------------------
+	postProcessWaitGroup.Wait()
```
whytheplatypus, line 126: with `StartProcessorWorkers` blocking and called here it can be wrapped in a `go func(){}` with a wait group declaired here. Similarly the webserver can also now be wrapped in a goroutine and tracked with that same top level wait group.

upload-server/internal/postprocessing/postprocessor.go
```diff
@@ -0,0 +1,34 @@
...
+	default:
+		for e := range c {
+			if err := Deliver(ctx, e.ID, e.Manifest, e.Target); err != nil {
+				go func(e Event) {
+					c <- e
+				}(e)
+			}
+			wg.Done()
+		}
+	}
```
whytheplatypus, line 33: not 100% sure this works, I think that `for range` is capturing until c is closed.. it may need to be

```go
for {
  select {
    case <-ctx.Done():
       // stuff
    case e <- c
      //stuff
  }
}
```

upload-server/testing/cli_test.go
```diff
@@ -324,7 +324,7 @@ func TestMain(m *testing.M) {
 		TusdHandlerBasePath:   "/files/",
 	}
 
-	handler, err := cli.Serve(appConfig)
+	handler, _, err := cli.Serve(context.Background(), appConfig)
```
whytheplatypus, line 327: with the changes suggested above this would now become a call to both Serve and StartProcessWorkers. That separation is a good thing.

whytheplatypus (commented):
Most of this review is around consolidating the wait groups rather than passing them around. Both work but passing them around makes it very difficult to track what's added too / declared done in a wait group.

whytheplatypus (changes requested):
sry hit the wrong button

upload-server/internal/postprocessing/postprocessor.go
```diff
@@ -0,0 +1,32 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case e := <-c:
+			if err := Deliver(ctx, e.ID, e.Manifest, e.Target); err != nil {
+				// TODO Retry
+				//go func(e Event) {
+				//	c <- e
+				//}(e)
```
whytheplatypus, line 28: Do something with the error here, logging is likely enough even if a bit redundant.

upload-server/testing/cli_test.go
```diff
@@ -324,7 +326,16 @@ func TestMain(m *testing.M) {
 		TusdHandlerBasePath:   "/files/",
 	}
 
-	handler, err := cli.Serve(appConfig)
+	testContext := context.Background()
+	var testWaitGroup sync.WaitGroup
```
whytheplatypus, line 330: add a `testWaitGroup.Wait()` somewhere just to complete the thought. Also this graceful shutdown behavior may eventually allow the tests to not have to use a timer to wait on reports to be written.

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #401, opened 2024-06-20
Upload 1395 ed3 n metadata smoke test additions

GSA2024:
Added ED3N manifests for Happy path and sad path scenarios.

### DIFF
--- a/tests/smoke/kotlin/src/test/resources/invalid_manifests_invalid_value.json
+++ b/tests/smoke/kotlin/src/test/resources/invalid_manifests_invalid_value.json
@@ -219,5 +219,39 @@
     "meta_ext_submissionperiod": "smoke submission period",
     "meta_ext_sourceversion": "V2022-12-31",
     "meta_username": "test-username"
+  },
+  {
+    "version": "2.0",
+    "data_stream_id": "ed3n",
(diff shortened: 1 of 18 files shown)


tests/smoke/kotlin/src/test/resources/invalid_manifests_invalid_value.json
```diff
@@ -219,5 +219,39 @@
...
+  },
+  {
+    "version": "2.0",
+    "data_stream_id": "ed3n",
+    "data_stream_route": "other",
+    "received_filename": "dex-smoke-test",
+    "jurisdiction": "blah",
+    "data_producer_id": "MA-NBS",
+    "sender_id": "ED3N-App",
+    "nbs_data_type": "MOLECULAR",
```
cfarmer-fearless, line 248: Maybe add one manifest that has mismatched character case to test the case sensitivity of the validation.  For example: "molecular" instead of "MOLECULAR"

GSA2024: Added invalid values for the fields

cfarmer-fearless (approved):

PR #417, opened 2024-07-05
Quickfix/deliver stream

cfarmer-fearless:
Refactor azure copy operation to use streams instead of async scheduler.  This is necessary because Azure is unable to perform a blob copy on a client's behalf if the source or destination storage accounts are in private vnets in their current configurations.  See this article for more details: https://learn.microsoft.com/en-us/troubleshoot/azure/azure-storage/blobs/connectivity/copy-blobs-between-storage-accounts-network-restriction#copy-blobs-between-storage-accounts-in-a-hub-spoke-architecture-using-private-endpoints

This patch uses workaround 4 specified in the article.  This is the same workaround that the upload processor (C# app) uses.  In the future,  we may want to consider other configurations so the router can take advantage of scheduled copying.

### DIFF
--- a/upload-server/internal/postprocessing/deliver.go
+++ b/upload-server/internal/postprocessing/deliver.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
 	"io"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/appconfig"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/health"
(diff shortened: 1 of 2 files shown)


upload-server/internal/postprocessing/deliver.go
```diff
@@ -172,29 +180,19 @@ func (ad *AzureDeliverer) Deliver(ctx context.Context, tuid string, manifest map
 	blobName, err := getDeliveredFilename(ctx, ad.Target, tuid, manifest)
 
 	destBlobClient := ad.ToContainerClient.NewBlobClient(blobName)
-	logger.Info("starting copy from", "src", srcBlobClient.URL(), "to dest", destBlobClient.URL())
-	resp, err := destBlobClient.StartCopyFromURL(ctx, srcBlobClient.URL(), nil)
+	s, err := srcBlobClient.DownloadStream(ctx, nil)
```
whytheplatypus, line 183: looks like `s.Body` is a read closer, add a `defer s.Body.Close()` just to be safe

whytheplatypus (changes requested):

upload-server/internal/postprocessing/deliver.go
cfarmer-fearless: Good catch, done

whytheplatypus (approved):

PR #420, opened 2024-07-08
UPLOAD-1352/update report structs to reflect schema changes

cfarmer-fearless:
These are new changes following the schema defined here: https://github.com/CDCgov/data-exchange-messages/blob/foundation/reports/reports/README.md.

These will be the new baseline for the report schemas.  Any changes to the schema going forward will be reflected in a new version.


upload-server/pkg/reports/builder.go
```diff
@@ -0,0 +1,213 @@
...
+			DataStreamRoute:   metadata.GetDataStreamRoute(b.Manifest),
+			Jurisdiction:      metadata.GetJurisdiction(b.Manifest),
+			DexIngestDatetime: metadata.GetDexIngestDatetime(b.Manifest),
+			ContentType:       "application/json",
+			DispositionType:   b.DispositionType,
+			StageInfo: ReportStageInfo{
+				Issues:           b.Issues,
+				Stage:            b.Stage,
+				Service:          "", // TODO get from version package
+				Version:          "", // TODO get from version package
```
whytheplatypus, line 183: finish this up and add repo

whytheplatypus (changes requested):
this looks great.

Because of the generic in Report, a `GetContentBuilder` or similar function could be make too, which would mean a writer wouldn't have to instantiate both the contentbuilder and the report builder... not sure if that's worth a change, just an idea.

upload-server/pkg/reports/builder.go
cfarmer-fearless: It's a dependency of this PR: https://github.com/CDCgov/data-exchange-upload/pull/419

cfarmer-fearless:
@whytheplatypus  I went ahead and removed content builder all together.  After reading your comment and looking at the code again, I decided it was redundant.  Thanks!

whytheplatypus (approved):

PR #427, opened 2024-07-17
Upload 1528/upload id in error response

cfarmer-fearless:
The purpose of this patch is to provide tus clients with the custom upload ID that the server creates when there are validation errors.  This is to allow these clients to gather PS API report information on that upload attempt more easily.

My initial approach to this was to simply concatenate the upload ID to the validation error message within the response body.  I then thought providing a structured json body with fields for the upload ID and validation errors was better because clients can parse this information out more easily.

### DIFF
--- a/upload-server/internal/metadata/metadata.go
+++ b/upload-server/internal/metadata/metadata.go
@@ -219,7 +219,7 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 				StatusCode: http.StatusBadRequest,
 				Body:       err.Error(),
 			})
-			return resp, nil
+			return resp, nil // Should return err here?
 		}
 		return resp, err
 	}
--- a/upload-server/internal/metadata/metadata.go
+++ b/upload-server/internal/metadata/metadata.go
@@ -233,7 +233,7 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 				StatusCode: http.StatusBadRequest,
 				Body:       string(b),
 			})
-			return resp, nil // Should return err here?
+			return resp, nil
 		}
 		return resp, err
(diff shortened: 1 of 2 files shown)


upload-server/internal/metadata/metadata.go
```diff
@@ -215,11 +220,20 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
...
+			if err != nil {
+				return resp, err
+			}
 			resp.HTTPResponse = resp.HTTPResponse.MergeWith(handler.HTTPResponse{
 				StatusCode: http.StatusBadRequest,
-				Body:       err.Error(),
+				Body:       string(b),
 			})
-			return resp, nil
+			return resp, nil // Should return err here?
```
whytheplatypus, line 236: you're good on not returning the err, i think that makes it 500.

upload-server/internal/metadata/metadata.go
```diff
@@ -215,11 +220,20 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
 
 		if errors.Is(err, validation.ErrFailure) {
 			resp.RejectUpload = true
+
+			respBody := PreCreateResponse{
+				UploadId:         tuid,
+				ValidationErrors: strings.Split(err.Error(), "\n"),
```
whytheplatypus, line 226: you might also be able to do `errors.Unwrap(err)` but i'm not sure how it would behave

whytheplatypus (approved):

upload-server/internal/metadata/metadata.go
cfarmer-fearless: I think would need to do `errors.Unwrap(err)` in a loop until it returns nil.

PR #428, opened 2024-07-18
Load Test post upload assertions

whytheplatypus:


tests/bad-uploader/assertions.go
```diff
@@ -0,0 +1,114 @@
...
+		for {
+			var errs error
+			select {
+			case <-timer.C:
+
+				client := graphql.NewClient(reportsURL, nil)
+
+				var q struct {
+					GetReports Reports `graphql:"getReports(uploadId: $id, reportsSortedBy: null, sortOrder: null)"`
+				}
```
whytheplatypus, line 64: should extract this.

tests/bad-uploader/assertions.go
```diff
@@ -0,0 +1,114 @@
...
+	resp, err := httpClient.Get(infoUrl)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to check upload: %d, %s", resp.StatusCode, infoUrl)
+	}
+	slog.Info("verified upload", "upload", infoUrl)
+
+	if reportsURL != "" {
```
whytheplatypus, line 53: this case can be extracted. Is there a way to decouple it from the global `reportsURL`?

tests/bad-uploader/assertions.go
```diff
@@ -0,0 +1,114 @@
...
+		timer := time.NewTicker(1 * time.Second)
```
whytheplatypus, line 54: make this generally configurable, probably doesn't have to be part of a TestCase

tests/bad-uploader/assertions.go
```diff
@@ -0,0 +1,114 @@
...
+				// If the file doesn't exist, create it, or append to the file
+				f, err := os.OpenFile(path.Base(upload)+".reports", os.O_CREATE|os.O_WRONLY, 0644)
+				if err != nil {
+					return err
+				}
+				defer f.Close()
+				je := json.NewEncoder(f)
+				if err := je.Encode(reports); err != nil {
+					return err
+				}
```
whytheplatypus, line 87: extract this and make it a type for different ways of recording the reports.

tests/bad-uploader/config.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+		slog.Error("Failed to convert env var, falling back to default", "error", err, "env", key)
+	}
+	return backup
+}
+func passthroughString(s string) (string, error) {
+	return s, nil
+}
+
+func init() {
```
whytheplatypus, line 196: move up

cfarmer-fearless:
Performs async checks for delivery info and PS API reports when report url and case file provided

tests/bad-uploader/main.go
```diff
@@ -2,490 +2,168 @@ package main
...
+			if limit == 0*time.Second {
+				limit = 1 * time.Minute
+			}
+			checkTimeout, cancel := context.WithTimeout(ctx, limit)
+			defer cancel()
+
+			wg.Add(len(PostUploadChecks))
+			for _, check := range PostUploadChecks {
+				check := check
+				go func(r *Result) {
```
whytheplatypus, line 81: Simplify by making `func(r *Result, check)`

tests/bad-uploader/info.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (ic *InfoChecker) DoCase(_ context.Context, c TestCase, uploadId string) error {
+	serverUrl, _ := path.Split(url)
+	infoUrl, err := neturl.JoinPath(serverUrl, "info", uploadId)
+	if err != nil {
+		return err
+	}
+
+	resp, err := ic.Client.Get(infoUrl)
```
whytheplatypus, line 52: use the context here with https://pkg.go.dev/net/http#NewRequestWithContext

tests/bad-uploader/main.go
```diff
@@ -2,490 +2,168 @@ package main
...
+					defer wg.Done()
+					// return a specific error and/or check result.  Specific error can have check specific info like upload id and reports
+					err := WithRetry(checkTimeout, r.testCase, uid, check.DoCase)
+					if err != nil {
+						slog.Error("failed post upload check", "error", err, "test case", r.testCase)
+						errs = errors.Join(errs, err)
+					} else {
+						check.OnSuccess()
```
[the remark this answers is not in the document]
cfarmer-fearless: The behavior might not be common though.  For example, for the info check we increment the delivery counter; for the event check, we increment the report counter.  Hence the interface method.

tests/bad-uploader/info.go
cfarmer-fearless: Done

cfarmer-fearless (approved):

PR #444, opened 2024-08-09
UPLOAD-1647/bugfix-event-handle-error-improvement

cfarmer-fearless:
This patch addresses an issue with the in-memory subscriber in which when an error occurred during the handling of an event, the go routine would hang because it would attempt to retry by putting the event back on the channel.  This is an unbuffered channel, so the go routine would block until something read the channel, but this would result in a deadlock because the parent go routine would wait for this go routine to complete before reading the channel again.  

A simple retry mechanism was implemented to mitigate this.  Events now contain a new field that holds a retry count.  When an error occurred during the handling of this event, the process will check if the count is less than some max number.  If that's true, it will put the event back on the channel but in a new go routine.  If it's false, it will just log the error and return.

### DIFF
--- a/upload-server/cmd/main.go
+++ b/upload-server/cmd/main.go
@@ -87,6 +87,7 @@ func main() {
 	logger.Info("starting app")
 
 	// Pub Sub
+	event.MaxRetries = appConfig.EventMaxRetryCount
 	// initialize event reporter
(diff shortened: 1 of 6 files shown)


rohitpanwar (approved):
LGTM

PR #445, opened 2024-08-12
add trailing slash for files with no date prefix

cfarmer-fearless:
This patch tweaks the logic for generating the upload file prefix by adding a trailing slash for all files and not just ones with the date prefix.


whytheplatypus (changes requested):
to make this a bit more robust update https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/internal/postprocessing/deliver.go#L349 to use [`filepath.Join`](https://pkg.go.dev/path/filepath#Join)

whytheplatypus (approved):

PR #447, opened 2024-08-14
UPLOAD-1582 Example metrics for uploads per data_stream

whytheplatypus:
Foundation work for displaying useful metrics based on the Manifest sent with a file. In this case completed uploads per data_stream.

example promethus query: `increase(files_uploaded_for_dextesting_count[1m])`

metrics:
- \# of responses by method and code
- \# of uploads by datastream
- \# of active uploads (might do by datastream?)


upload-server/cmd/cli/hooks.go
```diff
@@ -103,10 +107,40 @@ func PrebuiltHooks(ctx context.Context, appConfig appconfig.AppConfig) (tusHooks
...
 	return handler, nil
 }
+
+func ManifestCounter(event handler.HookEvent, resp tusHooks.HookResponse) (tusHooks.HookResponse, error) {
+	key := "data_stream_id"
+	val, ok := event.Upload.MetaData[key]
+	if !ok {
+		val, ok = resp.ChangeFileInfo.MetaData[key]
+		if !ok {
+			// no op.. should be something else
```
cfarmer-fearless, line 124: Assuming this would throw some kind of "invalid metadata" error.

cfarmer-fearless (commented):
Looks like a good start

whytheplatypus:
Testing example:
```
podman-compose up -d
# go to http://localhost:3000/d/eduxvq8up8f7kb/dex-dashboard?orgId=1&refresh=5s to see metrics, upload a file or two and watch them change
```

upload-server/grafana/provisioning/dashboards/dex.json
```diff
@@ -0,0 +1,358 @@
...
+          "legendFormat": "__auto",
+          "range": true,
+          "refId": "A",
+          "useBackend": false
+        }
+      ],
+      "title": "HTTP Responses",
+      "type": "piechart"
+    },
+    {
```
cyber-decker, line 92: This seems to be the start of the Active Uploads config.  Needs an appropriate target/expression to be used for displaying the results.

upload-server/grafana/provisioning/dashboards/dex.json
```diff
@@ -0,0 +1,358 @@
...
+          "includeNullMetadata": false,
+          "instant": false,
+          "legendFormat": "__auto",
+          "range": true,
+          "refId": "A",
+          "useBackend": false
+        }
+      ],
+      "title": "Uploads",
+      "type": "barchart"
```
cyber-decker, line 339: `timeseries` might work well here.  It was easier to visualize the change in upload counts over time with that.

cyber-decker (changes requested):
I was able to fire this up and launch the application just fine.  Looks like upload is working as expected.  I found the `demo-client-node-js` tool in `/upload-server/example/demo-client-node-js` to be really helpful here.

The metrics were getting reported correctly on the `localhost:8080/metrics` endpoint.  It looks like and things were mostly showing up on the dashboard but I did run in to a few things that might require some changes on the dashboard:

- Active Uploads was not correctly configured in grafana and was not showing any changes at all.  No metric was set for this dashboard tile. Might want to set this to `dex_server_active_uploads` with time series view.
- Uploads might be better as a time series graph.  The bar graph over time gives some weird, extrapolated metrics that don't really represent well what the graph is showing.   
- HTTP Responses is working fine, but, maybe something different than the pie chart.  The graph could get flooded with something like tons of patch requests for small chunks and you can't even see the rest of the graph unless you hide the patch responses.  This may be better as straight numbers or maybe bars?  Because the values can vary so differently, the percentage difference is too big when one thing is heavily inflated.

Other than that, the metrics seem to be getting into the system as expected, just need to tweak the display/output.

thetif (approved):
looks good to me
checked out the new metrics dashboard and it's updating correctly

upload-server/docker-compose.yml
```diff
@@ -1,35 +1,45 @@
...
+  prometheus:
+    image: prom/prometheus
+    container_name: prometheus
+    command:
+      - "--config.file=/etc/prometheus/prometheus.yml"
+    ports:
+      - 9090:9090
+    restart: unless-stopped
+    volumes:
+      - ./config/local/prometheus:/etc/prometheus
```
cyber-decker, line 12: There's nothing in the `./config/local` folders.  I think the things in `./configs/local` need to be moved in to there or change the target in the docker-compose configuration on this line. 

This is causing prometheus/grafana to get some errors on startup and not load the dex dashboards.

upload-server/docker-compose.yml
```diff
@@ -1,35 +1,45 @@
...
-      - "10000:10000"
-      - "10001:10001"
-      - "10002:10002"
+      - 3000:3000
+    restart: unless-stopped
+    environment:
+      - GF_SECURITY_ADMIN_USER=admin
+      - GF_SECURITY_ADMIN_PASSWORD=grafana
+    volumes:
+      - ./config/local/grafana/provisioning:/etc/grafana/provisioning
```
cyber-decker, line 24: Same problem as above.  needs to be either `./configs/local` or move the stuff from `./config/local` to this folder or change the target here in the docker-compose. 

Dashboards won't load up with this configuration.

cyber-decker (changes requested):
Needs one more little thing to finish here to get the docker compose configs right to point to the correct grafana/prometheus configs.   

I did that locally and it's looking good.  The dashboards looks great now and are reporting things nicely!

cyber-decker (approved):
Looks great!

PR #449, opened 2024-08-15
Upload 1658/s3 config loader

cfarmer-fearless:
Adds an implementation for loading an upload config from an S3 bucket.

TODO:
- Tidying.  There's some structural improvements we can make here for the secondary objects that depend on the environment the service is deployed in.  That includes the config loader, appender, locker, and router.  It would be great if we could initialize these objects in a more composable way, such as
```
hookHandler
  .withConfigLoader()
  .withMetadataAppender()
  .withRouter()
```
This is just an example.  Can also use option functions as that's a common pattern for composition in Go as well.

### DIFF
--- a/upload-server/internal/loaders/s3.go
+++ b/upload-server/internal/loaders/s3.go
@@ -1,0 +1,25 @@
+package loaders
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"io"
+)
+
+type S3ConfigLoader struct {
+	Client     *s3.Client
+	BucketName string
+}
+
+func (l *S3ConfigLoader) LoadConfig(ctx context.Context, path string) ([]byte, error) {
+	output, err := l.Client.GetObject(ctx, &s3.GetObjectInput{
+		Bucket: &l.BucketName,
+		Key:    &path,
+	})
+	if err != nil {
+
+		return nil, err
+	}
+
+	return io.ReadAll(output.Body)
+}
--- a/upload-server/cmd/cli/datastore.go
+++ b/upload-server/cmd/cli/datastore.go
@@ -3,8 +3,6 @@ package cli
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go-v2/config"
-	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/stores3"
 	"github.com/tus/tusd/v2/pkg/s3store"
 	"os"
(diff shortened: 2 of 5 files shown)


upload-server/internal/loaders/s3.go
```diff
@@ -0,0 +1,24 @@
...
+	"io"
+)
+
+type S3ConfigLoader struct {
+	Client     *s3.Client
+	BucketName string
+}
+
+func (l *S3ConfigLoader) LoadConfig(ctx context.Context, path string) ([]byte, error) {
+	output, err := l.Client.GetObject(ctx, &s3.GetObjectInput{
```
whytheplatypus, line 15: if output.Body is a `ReadCloser` which they often are add a `defer output.Body.Close()` for good measure

upload-server/internal/postprocessing/deliver.go
```diff
@@ -38,6 +38,46 @@ func init() {
...
+		routingDeliverer, err := NewAzureDeliverer(ctx, "routing", &appConfig)
+		if err != nil {
+			logger.Error("failed to connect to router deliverer target", "error", err.Error())
+		} else {
+			RegisterTarget("routing", routingDeliverer)
+			health.Register(routingDeliverer)
+		}
+	}
+
+	RegisterTarget("edav", edavDeliverer)
```
whytheplatypus, line 73: this could override the calls to `RegisterTarget` within the if statements. consider extracting the deliverer creation logic, that might clear things up.

upload-server/pkg/inspector/azure.go
```diff
@@ -1,4 +1,4 @@
-package azureinspector
+package inspector
```
whytheplatypus, line 1: consider leaving these as separate packages, could go either way but the separate packages is more in line with tusd patterns for example.

whytheplatypus (changes requested):

upload-server/pkg/inspector/azure.go
cfarmer-fearless: done

upload-server/internal/loaders/s3.go
cfarmer-fearless: done

upload-server/internal/postprocessing/deliver.go
cfarmer-fearless: Great catch.  I reworked it to only call `RegisterTarget` once at the end for each target, and only call `health.Register` once for each target if their cloud-based deliverer is not nil

whytheplatypus (approved):

PR #456, opened 2024-08-22
Playwright accessibility tests

cyber-decker:
Trying to build out some accessibility automation with playwright using the axe-core plugin.

- Added @axe-core/playwright package
- Created a manifest test against each manifest page to check for accessibility issues
- Created a landing page test for accessibility
- Created an accessibility test for the upload status page (default state)
- Added test parallelism (add a `--workers=4` command line param to run tests in parallel)
- Fixed some configuration for the upload UI port for all tests

Future things
- Improve the test failure reporting to make it easier to read through violations


tests/smoke/playwright/test/upload-test-accessibility.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const axeRuleTags = ["wcag2a", "wcag2aa", "wcag21a", "wcag21aa"];
+
+    test('Checks accessiblity for the upload landing page', async ({ page }, testInfo) => {
+        await page.goto(`/`)
+        const results = await new AxeBuilder({ page })
+            .withTags(axeRuleTags)
+            .analyze();
+
+        expect(results.violations).toEqual([]);
```
cfarmer-fearless, line 16: Use `toHaveLength` https://playwright.dev/docs/api/class-genericassertions#generic-assertions-to-have-length

tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -1,17 +1,36 @@
 import { expect, test } from '@playwright/test';
 
+test.describe.configure({ mode: 'parallel' });
+
 [
+    { dataStream: "celr", route: "csv" },
```
cfarmer-fearless, line 6: Could move this list to a separate ts file and export it, or export getters like `getAll` or `getByDataStream` to do some simple filtering

cfarmer-fearless (changes requested):

tests/smoke/playwright/test/upload-test-accessibility.spec.ts
cyber-decker: The disadvantage here would be that if the length is different, playwright will just tell us the violation length is different and not dump the list of violations.  This is a bit of a shortcut (maybe a hack?) that if the violations array isn't empty, it will show us the diff and output the violations list as well.

tests/smoke/playwright/test/upload-test-pages.spec.ts
cyber-decker: I thought about this too, but part of it abstracts away what the test is doing and makes it hard to know what's being tested.  I think there is room here to extend this object as well adding in specific expectations to check for, which might make your argument here more compelling because a really complex object as an input will make the test hard to read.  

Do you think we need to change this now or can we noodle over this a little and change it in a future story?

cfarmer-fearless (approved):

PR #457, opened 2024-08-23
feature/info-endpoint-delivery-status

cfarmer-fearless:


upload-server/cmd/cli/info.go
```diff
@@ -19,6 +19,7 @@ import (
 type UploadInspector interface {
 	InspectInfoFile(c context.Context, id string) (map[string]any, error)
 	InspectUploadedFile(c context.Context, id string) (map[string]any, error)
+	InspectFileStatus(ctx context.Context, id string) (*info.FileStatus, error)
```
whytheplatypus, line 22: consider putting this in a separate interface. There's likely to be a pattern where the uploaded file is checked with something like an s3 api and the status is gotten from another source (ps api, a different file system, etc.)

cfarmer-fearless: Good call.  Yeah was thinking composition would be better here.  Maybe have the upload inspector structs be composed of something that can check file status, as opposed to expanding the behavior of the upload inspector itself.

cfarmer-fearless: done

thetif (approved):

upload-server/pkg/azureinspector/azure.go
```diff
@@ -68,3 +68,8 @@ func (aui *AzureUploadInspector) InspectUploadedFile(c context.Context, id strin
 
 	return uploadedFileInfo, nil
 }
+
+func (aui *AzureUploadInspector) InspectFileStatus(ctx context.Context, id string) ([]info.FileDeliveryStatus, error) {
+	//TODO implement me
+	panic("implement me")
+}
```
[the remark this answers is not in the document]
cfarmer-fearless: Removed

upload-server/pkg/s3inspector/s3.go
```diff
@@ -68,3 +68,8 @@ func (sui *S3UploadInspector) InspectUploadedFile(c context.Context, id string)
 		"size_bytes": output.ObjectSize,
 	}, nil
 }
+
+func (sui *S3UploadInspector) InspectFileStatus(ctx context.Context, id string) ([]info.FileDeliveryStatus, error) {
+	//TODO implement me
+	panic("implement me")
+}
```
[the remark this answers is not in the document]
cfarmer-fearless: Removed

whytheplatypus (approved):

PR #464, opened 2024-08-28
File delivery retry tool Golang

cfarmer-fearless:
New replay tool that uses the route endpoint and is written in Golang.

Currently can
- Ingest a list of CSV files that contain a list of Upload IDs and their destination targets.
- Supported targets are routing and edav
- Issues POST request to internal server using private url.  No support for auth or public access.
- Retry in parallel using go routines

Future highlights
- Output a report for replay results
- Simple mode.  Take a simple list of Upload IDs and targets as opposed to CSV files


reupload-tool-go/main.go
```diff
@@ -0,0 +1,131 @@
...
+			})
+		}
+	}
+	slog.Info(fmt.Sprintf("replaying %d files", len(replays)))
+	slog.Debug(fmt.Sprintf("File IDs: %v", replays))
+
+	c := make(chan Replay, parallelism)
+	for i := 0; i < parallelism; i++ {
+		go worker(c)
+	}
```
whytheplatypus, line 96: move the wg logic here, the `for` loops in the workers will handle their lifecycle.

Can do something like `wg.Add(paralellism)` and then
```
go func(){
defer wg.Done()
worker(c)
}()
```

whytheplatypus (changes requested):

whytheplatypus (approved):

PR #468, opened 2024-09-03
Agnostic auth v1

ZacharyRizer:


upload-server/internal/appconfig/appconfig.go
```diff
@@ -53,6 +53,11 @@ type AppConfig struct {
 	LocalEDAVFolder       string `env:"LOCAL_EDAV_FOLDER, default=./uploads/edav"`
 	LocalRoutingFolder    string `env:"LOCAL_ROUTING_FOLDER, default=./uploads/routing"`
 
+	// OAUTH
+	OauthIntrospectionUrl string `env:"OAUTH_INTROSPECTION_URL, default=https://apigw-stg.cdc.gov:8443"`
```
cfarmer-fearless, line 57: Could group these into an Oauth config type struct and provide a prefix of `OAUTH_`

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,154 @@
...
+			if !validateOpaqueToken(ctx, w, token) {
+				return
+			}
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func validateJWT(ctx context.Context, w http.ResponseWriter, token string) bool {
```
cfarmer-fearless, line 47: These validation functions could be a bit more testable if they returned errors and didn't take a response writer.  The http handler can take care of the http response stuff based on the error

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,154 @@
...
+	// req.SetBasicAuth(appconfig.LoadedConfig.ClientID, appconfig.LoadedConfig.ClientSecret)
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil || resp.StatusCode != http.StatusOK {
+		http.Error(w, "Failed to validate opaque token", http.StatusUnauthorized)
+		return false
+	}
+	defer resp.Body.Close()
+
+	var introspectionResponse struct {
```
cfarmer-fearless, line 109: Define this struct at the package level

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,154 @@
...
+		Active bool   `json:"active"`
+		Scope  string `json:"scope"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&introspectionResponse); err != nil {
```
cfarmer-fearless, line 114: Use `json.Unmarshal` instead of `json.NewDecoder`

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return true
+}
+
+func hasRequiredScopes(actualScopes, requiredScopes []string) bool {
+	if len(requiredScopes) == 0 {
+		return true
+	}
+
+	scopeMap := make(map[string]bool)
```
cfarmer-fearless, line 144: Is the map necessary?  Can't you loop over actual and do `slice.Contains` on required?

ZacharyRizer: Probably not, been a bit since I used Go, I'll looking into optimizing.

upload-server/internal/appconfig/appconfig.go
ZacharyRizer: Good call, done.

upload-server/internal/middleware/authverification.go
ZacharyRizer: Good call, refactored.

upload-server/internal/middleware/authverification.go
ZacharyRizer: Done

upload-server/internal/middleware/authverification.go
ZacharyRizer: I would need to put the resp.Body into a byte array, using the io package I believe to use Unmarshal. Which is more preferable in your opinion?

upload-server/internal/middleware/authverification.go
ZacharyRizer: Simplified.

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,148 @@
...
+		token := authHeader[len("Bearer "):]
+
+		var err error
+		if strings.Count(token, ".") == 2 {
+			// Token is JWT, validate using oidc verifier
+			err = validateJWT(ctx, token)
+		} else {
+			// Token is opaque, validate using introspection
+			err = validateOpaqueToken(ctx, token)
+		}
```
whytheplatypus, line 48: consider changing this check to be based off what's configured: if an IssuerURL is provided validateJWT and return, if an Info endpoint is provided validateOpaqueToken and return.

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,148 @@
...
+	defer resp.Body.Close()
+
+	var introspectionResponse IntrospectionResponse
+
+	if err := json.NewDecoder(resp.Body).Decode(&introspectionResponse); err != nil {
+		return fmt.Errorf("failed to parse introspection response: %w", err)
+	}
+
+	if !introspectionResponse.Active {
+		return fmt.Errorf("inactive token")
```
whytheplatypus, line 120: create some constants for errors so that the middleware can react to them. (e.g. InactiveTokenErr -> 401)

upload-server/internal/middleware/authverification.go
ZacharyRizer: Where are you thinking the configuration will live? Currently the issuer is an env variable, and I was assuming we'd always have an issuer present. Same thing for the `endpoint`, are you thinking that we'd either have an issuer or introspection url? I was thinking we'd want it to be open to have both jwt's and opaque tokens be ingested at the same time.

upload-server/internal/middleware/authverification.go
ZacharyRizer: Updated

upload-server/readme.md
```diff
@@ -1,95 +1,159 @@
...
+- `REDIS_CONNECTION_STRING` - The full URI of the Redis instance, including authentication credentials such as username/password or access token. Make sure to use `rediss://` instead of `redis://` to use TLS for this traffic.
+
+### OAuth Token Verification Middleware
+
+- package location: `upload-server/internal/middleware/authverification.go`
+
+This middleware provides OAuth 2.0 token verification for incoming requests. It currently supports JWT tokens with the plan to add support for
+opaque tokens. You can use it to protect either your entire router or individual routes.
+
+##### Configuration
```
cfarmer-fearless, line 81: Use level 4 header instead of level 5

cfarmer-fearless (approved):
Small readme tweak but looks great!  Also wondering if the middleware function names can be improved to be a verb instead of a noun, like `VerifyOauth` or `WithOauthVerification`

upload-server/internal/middleware/authverification.go
```diff
@@ -0,0 +1,132 @@
...
+	return e.Msg
+}
+
+func NewHTTPError(code int, msg string) *HTTPError {
+	return &HTTPError{Code: code, Msg: msg}
+}
+
+func VerifyOAuthTokenMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		authEnabled := appconfig.LoadedConfig.OauthConfig.AuthEnabled
```
whytheplatypus, line 38: pull this dependency on appconfig out. So these handlers and functions can be methods on a struct that has things like `IssuerURL`

whytheplatypus (changes requested):
Wrap the tus and info handler in this.

upload-server/internal/middleware/authverification.go
ZacharyRizer: Refactored

whytheplatypus (approved):

PR #470, opened 2024-09-04
update H5 flu vax config

AJKimpo:
## Updates to the NDLP H5 Flu Vaccine Config
- updated config name to `h5-influenza-vaccination-csv.json`
- updated data_stream_id allowed value to `h5-influenza-vaccination`
- added custom metadata field meta_ext_sourceversion allowed value `V2022-12-31`
- removed all compatible field lines; this config was never a v1 config, so these lines were not needed
- added `XXA` allowed test value to jurisdiction
- added full list of allowed values to data_producer_id


edimeo (approved):
This looks a lot better. Approved

PR #471, opened 2024-09-04
H5 flu vax fix for allowed values

AJKimpo:
## Updates to H5 Flu vaccine config
- updated allowed values for data_stream_id and data_stream_route to be included in brackets, in conformance with other config files


edimeo (approved):
LGTM

PR #474, opened 2024-09-05
Automate graphql get reports functionalities

GSA2024:
Updated test "shouldHaveReportsForSuccessfulFileUpload: for PS-API graphql report validation, including metadata verification, 
transformation actions and upload status checks for the successful file upload.


tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -17,7 +17,8 @@ class ProcStat {
     private val testFile = TestFile.getResourceFile("10KB-test-file")
     private val authClient = DexUploadClient(EnvConfig.UPLOAD_URL)
     private val dexBlobClient = Azure.getBlobServiceClient(EnvConfig.DEX_STORAGE_CONNECTION_STRING)
-    private val procStatReqSpec = given().apply {
+    private val procStatReqSpec = given().relaxedHTTPSValidation()
+        .apply {
```
gulnaz-cdc, line 21: Check if using relaxedHTTPSValidation() is necessary, as it might lower security by skipping SSL checks. Make sure it's safe for the environment you're testing

cfarmer-fearless: Agree that it's less than ideal, but since we know this is hitting an environment only accessible from cdc network, it's ok to use.

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -43,56 +42,96 @@ class ProcStat {
...
+
+            val reportList = dataResponse.data.reports
+
+            reportList.forEach { report ->
+                val schemaName = report.content.contentSchemaName
+                when (schemaName) {
+                    "metadata-verify" -> {
+                        val uploadId = report.content.metadata?.uploadId
+                        val dexIngestDateTime = report.content.metadata?.dexIngestDateTime
+                        log.info("Metadata Verify - Upload ID: $uploadId, Dex Ingest DateTime: $dexIngestDateTime")
```
cfarmer-fearless, line 79: Assert that the upload id equals the uid var defined above.  
Assert that the metadata verify report status is successful and there are no isses.

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -43,56 +42,96 @@ class ProcStat {
...
+                    }
+
+                    "metadata-transform" -> {
+                        report.content.transforms?.forEach { transform ->
+                            val action = transform.action
+                            val field = transform.field
+                            val value = transform.value
+                            log.info("Transform Action: $action, Field: $field, Value: $value")
```
cfarmer-fearless, line 87: Assert that there is at least one transform within the content of this report

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -43,56 +42,96 @@ class ProcStat {
...
+                        }
+                    }
+
+                    "upload-started", "upload-completed" -> {
+                        val status = report.content.status
+                        assertEquals("SUCCESS", status, "Expected status 'SUCCESS' for $schemaName, but found: $status")
+                        log.info("Schema Name: $schemaName, Status: $status")
```
cfarmer-fearless, line 94: Many of these logs can be debug

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -43,56 +42,96 @@ class ProcStat {
...
+                    }
+
+                    "upload-status" -> {
+                        val filename = report.content.filename
+                        val tguid = report.content.tguid
+                        val offset = report.content.offset
+                        val size = report.content.size
+                        log.info("Upload Status - Filename: $filename, TGUID: $tguid, Offset: $offset, Size: $size")
```
cfarmer-fearless, line 102: Assert that the size and offset match

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -43,56 +42,96 @@ class ProcStat {
...
+                    }
+
+                    else -> {
+                        log.info("Unknown schema name: $schemaName")
+                    }
+                }
+            }
+
+        } catch (e: Exception) {
+            log.error("Test failed due to exception: ${e.message}", e)
```
cfarmer-fearless, line 133: This catch is probably unnecessary.  Just let the test fail instead of catching the exception

cfarmer-fearless (changes requested):

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
GSA2024: Added debug logs

cfarmer-fearless (approved):

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
cyber-decker: This is a common practice in test envionments since we are using self signed certs in the test envs.  This is normal and usual unless we are specifically confirming that we need valid certs in the test environments.  I feel like that's a different kind of test.

PR #501, opened 2024-09-20
Bugfix: Manifest Hydration Reports

cfarmer-fearless:
Small patch to fix the reports that get generated for manifest hydration.


rohitpanwar (approved):
LGTM

PR #503, opened 2024-09-23
day and month padding for path template

cfarmer-fearless:
Makes sure single digit days and months have a leading zero when using the path template.


thetif (approved):
looks good

PR #505, opened 2024-09-23
UPLOAD-1725 CSRF

thetif:
Added gorilla/csrf to prevent csrf


upload-server/internal/appconfig/appconfig.go
```diff
@@ -67,6 +68,7 @@ type AppConfig struct {
 	TusUIFileEndpointUrl string `env:"TUS_UI_FILE_ENDPOINT_URL, default=http://localhost:8080/files/"`
 	TusUIInfoEndpointUrl string `env:"TUS_UI_INFO_ENDPOINT_URL, default=http://localhost:8080/info/"`
 	UIPort               string `env:"UI_PORT, default=:8081"`
+	CsrfToken            string `env:"CSRF_TOKEN, default=SwVgY4SfiXNyXCT4U6AvLNURDYS7J+Y/V2j4ng2UVp0XwQY0IUELUT5J5b/FATcE"`
```
cfarmer-fearless, line 71: Ok to commit this default?

cfarmer-fearless (commented):
Looks pretty solid so far!

upload-server/internal/appconfig/appconfig.go
thetif: It's just a random string of 32 bytes. It's ok to commit it because it's only used for this local server, but I can also remove it, we would need to update the instructions to say that setting the CSRF_TOKEN is required.

cfarmer-fearless: Fine to leave then.  Add a comment saying it's for local use only and not sensitive.

upload-server/testing/cli_test.go
```diff
@@ -44,11 +44,21 @@ var (
 	}
 )
 
+const TestFolderUploadsTus = "uploads"
```
cfarmer-fearless, line 47: The tests use a `test` subdir instead of `uploads`

cfarmer-fearless (changes requested):

upload-server/testing/cli_test.go
thetif: yes, I duplicated the same structure in the tests, so it's pointing to the uploads dir inside the testing dir

upload-server/internal/delivery/file.go
```diff
@@ -99,3 +100,21 @@ func (fd *FileSource) GetMetadata(_ context.Context, tuid string) (map[string]st
 
 	return m, nil
 }
+
+func (fd *FileSource) Health(_ context.Context) (rsp models.ServiceHealthResp) {
+	rsp.Service = "File Source"
+	info, err := os.Stat(fd.Path)
```
whytheplatypus, line 106: consider using `fs.Stat(fd.FS, ".")` instead, fewer parts to coordinate

upload-server/internal/delivery/file.go
```diff
@@ -63,7 +63,8 @@ func (fd *FileDestination) Health(_ context.Context) (rsp models.ServiceHealthRe
 }
 
 type FileSource struct {
-	FS fs.FS
+	FS   fs.FS
+	Path string
```
whytheplatypus, line 67: Adds complexity for the health reporting.. if a name is needed maybe changing registering a health check to include a name is needed.

upload-server/internal/ui/ui.go
```diff
@@ -126,12 +149,13 @@ func GetRouter(uploadUrl string, infoUrl string) *http.ServeMux {
...
 			MetadataFields:  filterMetadataFields(config),
 			Navbar:          components.NewNavbar(false),
+			CsrfToken:       csrf.Token(r),
 		})
 		if err != nil {
 			http.Error(rw, err.Error(), http.StatusInternalServerError)
 			return
 		}
-	})
+	}).Methods("POST")
```
whytheplatypus, line 158: Should stay a GET so that it can be used as a bookmark to go directly to a form for the given datastream/route

whytheplatypus (changes requested):

upload-server/cmd/cli/serve.go
```diff
@@ -102,8 +101,12 @@ func Serve(ctx context.Context, appConfig appconfig.AppConfig) (http.Handler, er
 	http.Handle("/metrics", promhttp.Handler())
 
 	http.Handle("/info/{UploadID}", uploadInfoHandler)
+	http.Handle("/info/", uploadInfoHandler)
```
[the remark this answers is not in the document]
thetif: it was doing something weird before where it wasn't handling hitting the /info endpoint correctly and it was returning a 200 if there was no uploadId. That's not happening now so it was probably fixed in other parts of the code. Removing it.

upload-server/internal/ui/ui.go
thetif: We can't use the CSRF protection for the form without putting it in the post. Now that we have CSRF we cannot let them bookmark directly to the manifest form. That is essentially a request forgery because you aren't coming at that page from the intended location. We could switch it back to a GET but then we lose the CSRF restrictions on the index page. There is no way to do both though. I thought @AJKimpo had said we could just change the messaging and tell them they have to go to the index page first. I am willing to remove the CSRF from the index page, switch /manifest back to a GET, and whitelist it, if that's how we want to go.

upload-server/cmd/cli/serve.go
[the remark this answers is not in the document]
thetif: looked at this again and I was able to figure it out, the [docs](https://cs.opensource.google/go/go/+/go1.23.1:src/net/http/server.go;l=2505) say that the url path handling changed from Go 1.21 and 1.22 and now "/" is an implicit wildcard, so it will match anything not matched in the other paths. So if you hit `/info` or `/route` with no `UploadId` it would treat it like you hit the `/` endpoint and respond with a `200` and json like this
```json
{"system":"DEX","dex_product":"UPLOAD API","dex_app":"upload server","server_time":"2024-09-26T14:13:04-04:00"}
```
which the upload template was not to happy with. Now if you want `"/"` to only work for `/` you have to set it as `"/{$}"`. I changed it to this and now if you hit `/info` or `/route` it returns a 404 as you would expect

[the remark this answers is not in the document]
whytheplatypus: woah, awesome find, i didn't know they'd changed that under the hood!

whytheplatypus (approved):

upload-server/testing/cli_test.go
```diff
@@ -121,91 +189,59 @@ func TestTus(t *testing.T) {
...
-					t.Error("meta file was not copied to dex checkpoint for file", tuid)
-				}
-
-				if slices.Contains(config.Copy.Targets, "edav") {
-					// Check that the file exists in the target checkpoint folders.
-					if _, err := os.Stat("./test/edav/" + tuid); errors.Is(err, os.ErrNotExist) {
-						t.Error("file was not copied to edav checkpoint for file", tuid)
-					}
-
-					// Remove and re-route the file
```
cfarmer-fearless, line 135: Is this test for re-routing the file being preserved somewhere I'm not seeing?

upload-server/pkg/azureinspector/azure.go
```diff
@@ -54,8 +52,7 @@ func (aui *AzureUploadInspector) InspectInfoFile(c context.Context, id string) (
 }
 
 func (aui *AzureUploadInspector) InspectUploadedFile(c context.Context, id string) (map[string]any, error) {
-	filename := filepath.Join(aui.TusPrefix, id)
-	uploadBlobClient := aui.TusContainerClient.NewBlobClient(filename)
+	uploadBlobClient := aui.TusContainerClient.NewBlobClient(aui.TusPrefix + "/" + id)
```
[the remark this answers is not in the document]
thetif: I'm attempting to resolve this issue which might not really be an issue

CleanShot 2024-09-27 at 10 27 08

I saw that filepath.Join uses the os specific file path separator and because azure and aws use urls, I was wondering if using filepath.Join was causing an issue with interpreting the url correctly. But I haven't been able to confirm that yet, or track down what's actually happening

upload-server/testing/cli_test.go
thetif: yes, I moved it own into it's own test `TestRouteEndpoint` because it's seemed kind of heavy for doing on every test case, but if we need to test every case, I can move it back in

upload-server/internal/ui/ui.go
```diff
@@ -184,9 +197,10 @@ func GetRouter(uploadUrl string, infoUrl string) *http.ServeMux {
 		}
 
 		http.Redirect(rw, r, fmt.Sprintf("/status/%s", uuid.String()), http.StatusFound)
-	})
+	}).Methods("POST")
```
cfarmer-fearless, line 200: Should this be GET?

upload-server/testing/cli_test.go
```diff
@@ -121,91 +189,59 @@ func TestTus(t *testing.T) {
...
+			t.Error("could not create tuid")
+		} else {
+			// Check that the file exists in the target checkpoint folder
+			if _, err := os.Stat(TestEDAVFolder + "/" + tuid); errors.Is(err, os.ErrNotExist) {
+				t.Error("file was not copied to edav checkpoint for file", tuid)
 			}
 
-			t.Log("test case", name, "passed", tuid)
+			// Remove and re-route the file
+			// TODO probably best if this was in its own test function but this is at least good for happy path test for now
```
cfarmer-fearless, line 225: Can remove this TODO comment now

upload-server/testing/cli_test.go
cfarmer-fearless: Perfect!

cfarmer-fearless (approved):

PR #510, opened 2024-09-26
Quickfix/v1 mainfest translation delivery

cfarmer-fearless:
Handles edge case where file with v1 manifest is uploaded, and post finish hook tries to find upload config with same name but for v2.  If it doesn't find a v2 config, it falls back to v1.  If it doesn't find v1, it returns an error to the user.

For context, it tries to find a v2 config when a v1 manifest is used because the v1 manifest is hydrated with v2 manifest fields, including `verstion: 2.0`.  So when the metadata gets read in post finish, it appears as v2.

There are probably more long term fixes for this edge case, but we want to at some point soon remove this hydration behavior completely.

### DIFF
--- a/upload-server/clean.sh
+++ b/upload-server/clean.sh
@@ -11,4 +11,6 @@ rm -rf testing/test/uploads
 rm -rf testing/test/dex
 rm -rf testing/test/edav
 rm -rf testing/test/routing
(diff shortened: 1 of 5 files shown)


thetif (approved):
Looks good

whytheplatypus (approved):

PR #514, opened 2024-09-27
Upload 1744 playwright data stream route updates

cyber-decker:
Updating the data stream/routes that the playwright tests use to make sure they have the data stream and routes defined properly.  Previously I had defined a data stream and route like so:

```
        "dataStream": "covid",
        "route": "all-monthly-vaccination-csv"
```
when it should be...
```
        "dataStream": "covid-all-monthly-vaccination",
        "route": "csv"
```

The tests were working fine, but only because I was only testing the manifest page.  This does NOT allow for a successful upload using the datastream/route combination like that.

- Using a manifest.json to define the manifests in one place.
- Update tests to use the manifest.json instead of defined arrays of data stream/routes
- Extend the test that tests invalid manifest to show what should happen for a malformed data stream/route combination
- Update the end to end tests to use the manifest list to test all those combos.


tests/smoke/playwright/test/manifests.json
```diff
@@ -0,0 +1,42 @@
+[
+    {
+        "dataStream": "covid-all-monthly-vaccination",
+        "route": "csv"
+    },
+    {
+        "dataStream": "covid-bridge-vaccination",
```
AJKimpo, line 7: Please remove the covid-bridge-vaccination section. This data stream has been decommissioned. I'm removing configs and updating kotlin tests now.

AJKimpo (changes requested):

tests/smoke/playwright/test/manifests.json
AJKimpo: remove lines 6-8. the screen grab isolated the wrong lines.

tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -41,25 +32,32 @@ test.describe("Upload Manifest Page", () => {
             await expect(nextButton).toHaveText('Next')
         })
             
+    });
+        
+    [
+        { dataStream: 'invalid', route: 'invalid' },
+        // Need to fix the server to enable these tests, they are currently valid when they shouldn't be
```
cfarmer-fearless, line 39: Add link to jira ticket

tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -41,25 +32,32 @@ test.describe("Upload Manifest Page", () => {
...
+            const errorPagePromise = page.waitForResponse(`/manifest?data_stream_id=${dataStream}&data_stream_route=${route}`);
+    
+            await page.goto(`/`);
+            await page.getByLabel('Data Stream', { exact: true }).fill(dataStream);
+            await page.getByLabel('Data Stream Route').fill(route);
+            await page.getByRole('button', { name: /next/i }).click();
+            const errorPageResponse = await errorPagePromise
+    
+            await expect(errorPageResponse.status()).toBe(404)
+            await expect(page.locator('body')).toContainText(`open v2/${dataStream}-${route}.json: `)
```
cfarmer-fearless, line 53: We may just rerender the same form with an error message.  I'm thinking that's a better UX than showing a raw error message from the server, and having the user guess to click their browser back button to go back to the form and try again.

AJKimpo (approved):
lgtm

PR #515, opened 2024-09-30
Test branch

GSA2024:
1. Info - This test validates the file information, manifest and upload status, and verifies successful deliveries to the target location.

2. Health - Added health check dependent services and verifies the status for the list of services for a specific environment.
The environment (dev, tst, stg, prd) is extracted from the UPLOAD_URL. If the environment in the URL is not recognized, an exception is thrown.

3.  Added example local properties. Updated readMe doc that explains how to set up env vars or local props.


tests/smoke/kotlin/src/test/kotlin/util/EnvConfig.kt
```diff
@@ -11,6 +11,7 @@ class EnvConfig {
         } else null
         val UPLOAD_URL: String = properties?.getProperty("upload.url") ?: System.getenv("UPLOAD_URL")
         val PROC_STAT_URL: String = properties?.getProperty("ps.api.url") ?: System.getenv("PS_API_URL")
+        val INFO_URL: String = properties?.getProperty("info.url") ?: System.getenv("INFO_URL")
         val SAMS_USERNAME: String = properties?.getProperty("sams.username") ?: System.getenv("SAMS_USERNAME")
         val SAMS_PASSWORD: String = properties?.getProperty("sams.password") ?: System.getenv("SAMS_PASSWORD")
         val DEX_STORAGE_CONNECTION_STRING: String = properties?.getProperty("dex.storage.connection.string") ?: System.getenv("DEX_STORAGE_CONNECTION_STRING")
```
cyber-decker, line 17: I think we need to put this in the documentation to include what properties need to go into the local.properties file and maybe even where they can be found and perhaps a little bit about what they do in this test suite. I can't run this until I track down the INFO_URL that has been added in

tests/smoke/kotlin/src/test/kotlin/Info.kt
```diff
@@ -43,23 +41,36 @@ class Info {
...
             }
         }
-    }
 
-    @Test(
-        groups = [Constants.Groups.FILE_INFO],
-        expectedExceptions = [IOException::class],
-        expectedExceptionsMessageRegExp = "Error getting file info.*"
-    )
-    fun shouldReturnNotFoundGivenInvalidId() {
```
cfarmer-fearless, line 62: Why was this test removed?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
```diff
@@ -2,23 +2,25 @@ package dex
...
 import java.nio.charset.StandardCharsets
+import util.EnvConfig
+import org.slf4j.LoggerFactory
 
 class DexUploadClient(private val url: String) {
     private val httpClient = OkHttpClient()
     private val objectMapper = ObjectMapper().registerModule(KotlinModule.Builder().build())
+    private val logger = LoggerFactory.getLogger(DexUploadClient::class.java)
 
     fun getToken(username: String, password: String): String {
```
cfarmer-fearless, line 22: Can you explain the changes to the `getToken` function?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
```diff
@@ -28,38 +30,39 @@ class DexUploadClient(private val url: String) {
...
+            val authResponse: AuthResponse = objectMapper.readValue(responseBody)
+            return authResponse.accessToken
+        }
     }
 
     fun getFileInfo(id: String, authToken: String): InfoResponse {
+
         val req = Request.Builder()
-            .url("$url/upload/info/$id")
+            .url("${EnvConfig.INFO_URL}/info/$id")
```
cfarmer-fearless, line 49: Shouldn't need a separate config variable for the info url.  Should be able to use the upload url appended with "/info".  Why do we need this new config variable?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
```diff
@@ -28,38 +30,39 @@ class DexUploadClient(private val url: String) {
...
             .header("Authorization", "Bearer $authToken")
             .build()
 
-        val resp = httpClient
-            .newCall(req)
-            .execute()
+        Thread.sleep(1000) // wait to load delivery response
```
cfarmer-fearless, line 53: What's the reason for this wait?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
```diff
@@ -28,38 +30,39 @@ class DexUploadClient(private val url: String) {
...
+            if (!response.isSuccessful) {
+                throw IOException("Error getting file info. ${response.message}")
+            }
 
-        val respBody: InfoResponse = objectMapper.readValue(resp.body?.string()
-            ?: throw IOException("Empty response"), InfoResponse::class.java)
+            val responseBody = response.body?.string() ?: throw IOException("Empty response body from server")
 
-        return respBody
+            logger.info("Raw response body: $responseBody")
```
cfarmer-fearless, line 62: Can be debug level log

cfarmer-fearless: Also can be a more specific message.  Raw response body from which request?  Which service?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
```diff
@@ -68,17 +71,15 @@ class DexUploadClient(private val url: String) {
             .header("Authorization", "Bearer $authToken")
             .build()
 
-        val resp = httpClient
-            .newCall(req)
-            .execute()
+        httpClient.newCall(req).execute().use { response ->
```
cfarmer-fearless, line 74: Any benefit of handling the response with `use` instead of just storing it in a local var?

cfarmer-fearless (changes requested):

tests/smoke/kotlin/src/test/kotlin/Health.kt
```diff
@@ -20,10 +20,15 @@ class Health {
 
         val expectedDependentServices = arrayOf(
             "Event Publishing processing-status-cosmos-db-report-sink-topics",
+            "Event Publishing ocio-ede-stg-upload-file-ready-topic",
+            "ocio-ede-stg-upload-file-ready-subscription Event Subscriber",
```
cyber-decker, line 24: These two are probably not going to be valid for other environments and are causing failures when I run this in test.  

This causes the assertion on line 44 to fail but it doesn't say why.

tests/smoke/kotlin/src/test/kotlin/util/EnvConfig.kt
GSA2024: I agree. 
I check with Chase, there is no need of changing the endpoint url for info and health check, need to update back> If possible, I can check-in local properties too if everyone is ok

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: Removed EnvConfig for info endpoint url, reverted upload url appended for info as-is

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: Logger used for debugging purposes. reverted the 'getToken' changes as it is not needed

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: `use` is safer, reduces the risk of resource leakage, and results in more concise and readable code. 

Since we have reverted changes as it not needed for both the methods `getToken` and `getFileInfo`

tests/smoke/kotlin/src/test/kotlin/Health.kt
GSA2024: Can I raise different PR/ticket for health check changes?

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: Removed

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: Removed Raw response

tests/smoke/kotlin/src/test/kotlin/Info.kt
GSA2024: Adding it back

tests/smoke/kotlin/src/test/kotlin/dex/DexUploadClient.kt
GSA2024: Logger is added for debugging purposes.

tests/smoke/kotlin/src/test/kotlin/util/EnvConfig.kt
cyber-decker: I still think we should include this in documentation.  This was a very challenging point of getting these tests running.  Documenting what all of these properties are used for and where they should be set is necessary.

tests/smoke/kotlin/src/test/kotlin/Health.kt
GSA2024: Just added additional assertions to compare the list of actual services with a list of expected services, which checks for matching elements without considering order. If the lists don't match, a detailed error message is generated.

GSA2024: Shall we add these two environment specific dependent services in EnvConfig file?
 "Event Publishing ocio-ede-tst-upload-file-ready-topic",
 "ocio-ede-tst-upload-file-ready-subscription Event Subscriber"
 @cfarmer-fearless @cyber-decker

cyber-decker: This seems complex. What would have to happen with other environments?  How would someone know what this should be changed to for any other environment this exists in?

cyber-decker:
When running the latest suite of tests, I currently have 62 tests passed and 7 failed tests.  None of them seem to be info related, but there is a health check related test that has not passed and some others that might need to grab the latest changes from main to address some datastream/route changes that have been made.

tests/smoke/kotlin/src/test/kotlin/util/EnvConfig.kt
GSA2024: Shall I put local properties in readMedoc?
@cfarmer-fearless @cyber-decker

tests/smoke/kotlin/src/test/kotlin/Health.kt
GSA2024: Fix tests for Health check - Added health check dependent services and verifies the status for the list of services for a specific environment.

The environment (dev, tst, stg, prd) is extracted from the UPLOAD_URL. If the environment in the URL is not recognized, an exception is thrown.

cyber-decker (approved):
I think this looks good for now.  We should definitely keep iterating over updating the documentation for these tests.

tests/smoke/kotlin/src/test/kotlin/Info.kt
GSA2024: It was removed accidentally it seems, `shouldReturnNotFoundGivenInvalidId` added back

cfarmer-fearless (approved):

PR #520, opened 2024-10-02
UPLOAD-1740 updating the README

thetif:
- moved the `.env.example` file to `/upload-server` so that it could be used by the docker configs
- added all environment variables to the `.env.example` and added comments to clarify their uses to the user
- cleaned up code where necessary


upload-server/internal/appconfig/appconfig.go
```diff
@@ -43,74 +43,88 @@ type AppConfig struct {
...
+	// Local file system delivery config
+	LocalRoutingFolder string `env:"LOCAL_ROUTING_FOLDER, default=./uploads/routing"`
+	LocalEDAVFolder    string `env:"LOCAL_EDAV_FOLDER, default=./uploads/edav"`
+	LocalEhdiFolder    string `env:"LOCAL_EHDI_FOLDER, default=./uploads/ehdi"`
+	LocalEicrFolder    string `env:"LOCAL_EICR_FOLDER, default=./uploads/eicr"`
+	LocalNcirdFolder   string `env:"LOCAL_NCIRD_FOLDER, default=./uploads/ncird"`
 
-	// UI
-	TusUIFileEndpointUrl string `env:"TUS_UI_FILE_ENDPOINT_URL, default=http://localhost:8080/files/"`
-	TusUIInfoEndpointUrl string `env:"TUS_UI_INFO_ENDPOINT_URL, default=http://localhost:8080/info/"`
```
thetif, line 70: Added `SERVER_PROTOCOL`, `SERVER_HOSTNAME`, and `TUSD_HANDLER_INFO_PATH` so that we could remove `TUS_UI_FILE_ENDPOINT_URL` and `TUS_UI_INFO_ENDPOINT_URL` because they were dependent on SERVER_PORT in a way that might not be intuitive to the user

upload-server/.env.example
```diff
@@ -0,0 +1,192 @@
+########################
```
thetif, line 1: Added all the possible environment variables here, grouped them by use, and added comments to make things clearer

upload-server/Dockerfile
```diff
@@ -36,5 +36,4 @@ RUN chown app:app ./dextusd
 USER app
 
 EXPOSE 8080
-
-CMD ["./dextusd"]
+ENTRYPOINT [ "./dextusd" ]
```
thetif, line 39: Switched from `CMD` to `ENTRYPOINT` so that we can pass in addition flags when running the container

upload-server/cmd/cli/flags.go
```diff
@@ -6,14 +6,12 @@ import (
 
 var Flags struct {
 	AppConfigPath string // if override
-	FileHooksDir  string
```
thetif, line 9: We're not actually using this flag anywhere so I removed it

upload-server/cmd/main.go
```diff
@@ -27,6 +27,14 @@ import (
 const appMainExitCode = 1
 
 func init() {
+	// ------------------------------------------------------------------
```
thetif, line 30: not sure if we were intentionally removing the flags or if it was removed by mistake, but added it back so that we could pass in the `.env` file for the configs

alternatively, we could have it use a `.env` by default if there is one present in the same directory

upload-server/internal/appconfig/appconfig.go
```diff
@@ -43,74 +43,88 @@ type AppConfig struct {
...
-	RoutingCheckpointContainer string `env:"ROUTING_CHECKPOINT_CONTAINER_NAME, default=routing-checkpoint"`
-	EhdiCheckpointContainer    string `env:"EHDI_CHECKPOINT_CONTAINER_NAME, default=ehdi-checkpoint"`
-	EicrCheckpointContainer    string `env:"EICR_CHECKPOINT_CONTAINER_NAME, default=eicr-checkpoint"`
-	NcirdCheckpointContainer   string `env:"NCIRD_CHECKPOINT_CONTAINER_NAME, default=ncird-checkpoint"`
 
-	Metrics MetricsConfig `env:", prefix=METRICS_"`
+	// S3 delivery config
+	RoutingS3Connection *S3StorageConfig `env:", prefix=ROUTING_S3_, noinit"`
+	EdavS3Connection    *S3StorageConfig `env:", prefix=EDAV_S3_, noinit"`
+	EhdiS3Connection    *S3StorageConfig `env:", prefix=EHDI_S3_, noinit"`
```
thetif, line 125: Added the missing destination types

upload-server/internal/appconfig/appconfig.go
```diff
@@ -276,11 +290,25 @@ func ParseConfig(ctx context.Context) (AppConfig, error) {
 	if err := envconfig.Process(ctx, &ac); err != nil {
 		return AppConfig{}, err
 	} // .if
+
 	if ac.AzureConnection != nil {
-		if ac.AzureConnection.ContainerEndpoint == "" {
+		if ac.AzureConnection.StorageName == "" || ac.AzureConnection.StorageKey == "" {
```
thetif, line 295: if only the AZURE_STORAGE_KEY is set, then AzureConnection is not nil and it tries to start up but fails. Added checks to see if both values we need are set and if not then I set AzureConnection to nil

upload-server/internal/appconfig/appconfig.go
```diff
@@ -276,11 +290,25 @@ func ParseConfig(ctx context.Context) (AppConfig, error) {
...
+			ac.AzureConnection = nil
+		} else if ac.AzureConnection.ContainerEndpoint == "" {
 			ac.AzureConnection.ContainerEndpoint = fmt.Sprintf("https://%s.blob.core.windows.net", ac.AzureConnection.StorageName)
 		}
 	}
+
+	if ac.S3Connection != nil {
```
thetif, line 302: similar to the AzureConnection

MegMurray (approved):
Looks great, Tif!

AJKimpo (approved):
This all looks good from my perspective; please have Chase review prior to merging

upload-server/.env.example
```diff
@@ -0,0 +1,188 @@
...
+TUS_AZURE_CONTAINER_NAME=
+DEX_MANIFEST_CONFIG_CONTAINER_NAME=
+
+# S3 upload bucket
+S3_ENDPOINT=
+S3_BUCKET_NAME=
+AWS_REGION=
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
+DEX_MANIFEST_CONFIG_BUCKET_NAME=
```
cfarmer-fearless, line 61: Add a comment saying only one of the bucket or the folder env vars are necessary, not both

upload-server/.env.example
```diff
@@ -0,0 +1,188 @@
...
+###########################################
+#### File delivery target configs ####
+###########################################
+
+#### Routing delivery target ####
+
+# local file system routing directory
+LOCAL_ROUTING_FOLDER= # has default
+
+# Azure routing container
```
cfarmer-fearless, line 101: These routing guys can be removed

upload-server/internal/appconfig/appconfig.go
cfarmer-fearless: Let's fail a bit louder here.  If there isn't enough info to connect, the app should fail to start.  Same for S3

upload-server/readme.md
```diff
@@ -2,7 +2,20 @@
...
+## Overview
+
+The DEX Upload API is an open-source tool that allows users to upload and manage data sets for public health initiatives. It is designed for ease of use and customization while also ensuring compliance with federal standards.
+
+The DEX Upload API is built on the [tus](https://tus.io) open protocol, which provides a way to upload large files in smaller chunks over HTTP. Each file chunk is sent as an HTTP PATCH request. The last PATCH request tells tus to combine the chunks into the full file. Once the file upload is complete, the server can also push it out to other destinations.
+
+### Key Features
+
+- Resumable uploads
+- Support for multiple formats (JSON, XML)
```
cfarmer-fearless, line 14: This bullet point may needs to be expanded a bit.  Multiple formats for what?

cfarmer-fearless (changes requested):
Some minor changes but looks great overall

upload-server/readme.md
thetif: I stole that part from @MegMurray. :D I guess multiple file formats since we aren't really restricting anything

upload-server/readme.md
```diff
@@ -11,218 +24,558 @@ Repo is structured (as feasible) based on the [golang-standards/project-layout](
...
+
+```shell
+source env-file
+go run ./cmd/main.go
+```
+
+If you name this file `.env` you can get the benefits of the [dotenv file format](https://www.dotenv.org/docs/security/env). For instance, it will automatically be recognized and loaded by tools like `docker-compose`. The `.env.example` file in the `upload-server/` directory contains all of the available environment variables for configuring the system. Add any environment variables you would like to set to your `.env` file.
+
+>[!WARNING]
+> Never check your `.env` file into source control. It should only be on your local computer or on the server you are using it on.
```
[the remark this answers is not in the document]
thetif: there is, I just wanted to make sure no one removed it because they thought they came up with a great idea to share their passwords with the team

upload-server/.env.example
```diff
@@ -0,0 +1,188 @@
...
+REDIS_CONNECTION_STRING=
+
+# oauth token verification middleware
+OAUTH_AUTH_ENABLED= # has default
+OAUTH_INTROSPECTION_URL=
+OAUTH_ISSUER_URL=
+OAUTH_REQUIRED_SCOPES=
+
+# processing status health
+PROCESSING_STATUS_HEALTH_URI=
```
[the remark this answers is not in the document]
thetif: cool, I wasn't sure what that was for but I didn't want to removed it if we needed it

upload-server/internal/appconfig/appconfig.go
```diff
@@ -292,15 +292,15 @@ func ParseConfig(ctx context.Context) (AppConfig, error) {
 
 	if ac.AzureConnection != nil {
 		if ac.AzureConnection.StorageName == "" || ac.AzureConnection.StorageKey == "" {
-			ac.AzureConnection = nil
+			return AppConfig{}, fmt.Errorf("missing required values for connecting to Azure")
```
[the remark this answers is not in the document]
thetif: I tried but Go did no like that, I think I could if we were returning a pointer to the config instead of the whole config

upload-server/readme.md
```diff
@@ -11,8 +11,9 @@ The DEX Upload API is built on the [tus](https://tus.io) open protocol, which pr
 ### Key Features
 
 - Resumable uploads
-- Support for multiple formats (JSON, XML)
+- Support for multiple file formats (e.g., JSON, XML)
```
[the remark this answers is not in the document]
thetif: yes, I'll add uploads to that

upload-server/cmd/main_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	dexTesting "github.com/cdcgov/data-exchange-upload/upload-server/testing"
 )
 
-var AZURITE_KEY = os.Getenv("AZURITE_STORAGE_KEY")
+var AZURITE_KEY = os.Getenv("AZURE_STORAGE_KEY")
```
whytheplatypus, line 19: suggest leaving this as "AZURITE_STORAGE_KEY" to separate it for testing.

upload-server/docker-compose.azurite.yml
```diff
@@ -1,21 +1,20 @@
...
   upload-server:
     environment:
-      - AZURE_STORAGE_ACCOUNT=devstoreaccount1
+      - AZURE_STORAGE_ACCOUNT=${AZURE_STORAGE_ACCOUNT:-devstoreaccount1}
       # Default test storage key can be found here https://github.com/Azure/Azurite?tab=readme-ov-file#default-storage-account
       - AZURE_STORAGE_KEY=${AZURE_STORAGE_KEY}
-      - AZURE_ENDPOINT=http://azurite:10000/devstoreaccount1
-      - TUS_AZURE_CONTAINER_NAME=test
+      - AZURE_ENDPOINT=${AZURE_ENDPOINT:-http://azurite:10000/devstoreaccount1}
+      - TUS_AZURE_CONTAINER_NAME=${TUS_AZURE_CONTAINER_NAME:-test}
```
whytheplatypus, line 17: is the usecase for these env overrides for if you want to run this upload-server config without azurite?

upload-server/internal/appconfig/appconfig.go
```diff
@@ -276,11 +289,25 @@ func ParseConfig(ctx context.Context) (AppConfig, error) {
 	if err := envconfig.Process(ctx, &ac); err != nil {
 		return AppConfig{}, err
 	} // .if
+
 	if ac.AzureConnection != nil {
-		if ac.AzureConnection.ContainerEndpoint == "" {
+		if ac.AzureConnection.StorageName == "" || ac.AzureConnection.StorageKey == "" {
+			return AppConfig{}, fmt.Errorf("missing required values for connecting to Azure")
+		} else if ac.AzureConnection.ContainerEndpoint == "" {
```
whytheplatypus, line 296:
```suggestion
		} 
		if ac.AzureConnection.ContainerEndpoint == "" {
```

shouldn't need the else because of the early return.

upload-server/testing/cli_test.go
```diff
@@ -629,9 +628,7 @@ func TestMain(m *testing.M) {
 	ts = httptest.NewServer(serveHandler)
 
 	// Start ui server
-	appConfig.TusUIFileEndpointUrl = ts.URL + "/files"
-	appConfig.TusUIInfoEndpointUrl = ts.URL + "/info"
-	uiHandler := ui.GetRouter(appConfig.TusUIFileEndpointUrl, appConfig.TusUIInfoEndpointUrl)
+	uiHandler := ui.GetRouter(ts.URL+appConfig.TusdHandlerBasePath, ts.URL+appConfig.TusdHandlerInfoPath)
```
whytheplatypus, line 631: it looks like these values will only be populated if `ParseConfig` is called

whytheplatypus (changes requested):

upload-server/testing/cli_test.go
thetif: maybe I'm miss understanding how appconfig.go works, but I don't think that's the case, it seems to create it fine without the flag

whytheplatypus: yup ignore me i was look at this and misread url vs path https://github.com/CDCgov/data-exchange-upload/pull/520/files#diff-81f4a76d0fca50bb683250ae015cebfbfd13708b4a189a04f9609015aa71310fR308-R309

upload-server/docker-compose.azurite.yml
thetif: yes, but I guess for this and minio we don't really need to have those values be configurable. I'll switch those back

thetif: I was mostly changing things so that the ports would be configurable in the docker-compose and then I think I got carried away

whytheplatypus (approved):

cyber-decker (approved):
Changes look great!  Thanks for the clarity edits on the env-configs readme!  Much much improved.

cfarmer-fearless (approved):

PR #522, opened 2024-10-04
UPLOAD-1710/UploadReports

ZacharyRizer:


upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+			if err != nil {
+				log.Printf("Error fetching data from GraphQL API: %v", err)
+				return
+			}
+
+			dataChan <- rowData
+		}(ds)
+	}
+
+	wg.Wait()
```
cfarmer-fearless, line 132: You probably don't even need the `Wait` since the main thread will block on the `range` over the channel.  If you must keep it, move it to right before the `return`.  You want the main thread to block on the `range` statement over the channel

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+	writer.Flush()
+
+	if err := writer.Error(); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+func saveCsvToFile(csvData []byte) error {
```
cfarmer-fearless, line 169: Use `io.Writer` instead of `[]byte`.  The data will probably only be `[]byte` but setting the type as `io.Writer` passes a pointer instead of a copy of all the bytes

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+	workingDir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get the current working directory: %v", err)
+	}
+
+	fullPath := filepath.Join(workingDir, "upload-report.csv")
```
cfarmer-fearless, line 175: Create a command line argument for the output path and default it to `os.Getwd`

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+
+	file, err := os.Create(fullPath)
```
cfarmer-fearless, line 177: Do you need to check if the path already exists first?

upload-reports/utils/config.go
```diff
@@ -0,0 +1,53 @@
...
+	Endpoint   string
+}
+
+func GetEnvVar(key string) string {
+	val := os.Getenv(key)
+	return val
+}
+
+func GetConfig() AppConfig {
+	dataStreams := (GetEnvVar("DATASTREAMS"))
```
cfarmer-fearless, line 27: dataStreams, startDate, endDate, and targetEnv should all be command line args.  See https://pkg.go.dev/flag.  You can also default the date range to some small range like the last 24 hours, and default targetEnv to "dev".

upload-reports/utils/config.go
```diff
@@ -0,0 +1,53 @@
...
+	startDate := GetEnvVar("START_DATE")
+	endDate := GetEnvVar("END_DATE")
+	targetEnv := GetEnvVar("TARGET_ENV")
```
cfarmer-fearless, line 30: Suggest rename targetEnv to reportFilenameSuffix since you could really put any value in this field.  For example, we may want to put the hostname of the environment like "api.cdc.gov" instead of just "prd".

upload-reports/utils/config.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func GetEnvVar(key string) string {
+	val := os.Getenv(key)
+	return val
+}
+
+func GetConfig() AppConfig {
+	dataStreams := (GetEnvVar("DATASTREAMS"))
+	startDate := GetEnvVar("START_DATE")
```
cfarmer-fearless, line 28: Add a command line arg for pointing to a `.env` file and see https://github.com/joho/godotenv

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+		if endpoint != "" {
+			o.UsePathStyle = true
+			o.BaseEndpoint = &endpoint
+		}
+	})
+
+	return client, nil
+}
+
+func uploadCsvToS3(ctx context.Context, client *s3.Client, bucketName string, key string, csvData []byte) error {
```
cfarmer-fearless, line 208: Change the type of `csvData` to `io.Reader` and pass `bytes.NewReader(csvData)` for that argument.

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+			strconv.FormatInt(rowData.UploadCount, 10),
+			strconv.FormatInt(rowData.DeliverySuccessCount, 10),
+			strconv.FormatInt(rowData.DeliveryEndCount, 10),
+		})
+	}
+
+	return csvData
+}
+
+func createCSV(data [][]string) ([]byte, error) {
```
cfarmer-fearless, line 150: Return an `io.Writer` instead of `[]byte`

cfarmer-fearless (approved):
Nice work on getting this functional.  Here is some feedback for making the code a bit more idiomatic.  Also, please include a readme with some usage examples and description of required environment variables.

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func fetchDataForDataStream(apiURL string, datastream string, route string, startDate string, endDate string) (ReportDataRow, error) {
+	ctx := context.Background()
+	client := graphql.NewClient(apiURL, http.DefaultClient)
+
+	resp, err := psApi.GetUploadStats(ctx, client, datastream, route, startDate, endDate)
+
+	if err != nil {
+		fmt.Printf("There was an issue reaching graphql: %v\n", err)
```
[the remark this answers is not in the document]
ZacharyRizer: Done

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+			log.Fatalf("failed to create S3 client: %v", err)
+		}
+
+		if err := uploadCsvToS3(ctx, client, config.S3Config.BucketName, key, csvBytes); err != nil {
+			log.Fatalf("Error uploading CSV to S3: %v", err)
+		}
+	}
+}
+
+func fetchDataForDataStream(apiURL string, datastream string, route string, startDate string, endDate string) (ReportDataRow, error) {
```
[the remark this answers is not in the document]
ZacharyRizer: Done

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	return reportRow, nil
+}
+
+func getCsvData(config utils.AppConfig) [][]string {
+	var csvData [][]string
+	csvData = append(csvData, []string{"Data Stream", "Route", "Start Date", "End Date", "Upload Count", "Delivery Success Count", "Delivery Fail Count"})
+
+	cleanedStartDate, err := utils.FormatDateString(config.StartDate)
```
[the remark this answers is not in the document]
ZacharyRizer: Split inputs out from the config to cleaned up values.

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		log.Fatalf("End date is in incorrect format: %v", err)
+	}
+
+	datastreams := strings.Split(config.DataStreams, ",")
+	var wg sync.WaitGroup
+	dataChan := make(chan ReportDataRow, len(datastreams))
+
+	for _, ds := range datastreams {
+		wg.Add(1)
```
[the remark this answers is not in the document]
ZacharyRizer: Done

upload-reports/main.go
```diff
@@ -0,0 +1,226 @@
...
+		DeliverySuccessCount: resp.GetGetUploadStats().PendingUploads.TotalCount,
+		DeliveryEndCount:     resp.GetGetUploadStats().UnDeliveredUploads.TotalCount,
+	}
+
+	return reportRow, nil
+}
+
+func getCsvData(config utils.AppConfig) [][]string {
+	var csvData [][]string
+	csvData = append(csvData, []string{"Data Stream", "Route", "Start Date", "End Date", "Upload Count", "Delivery Success Count", "Delivery Fail Count"})
```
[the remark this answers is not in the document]
ZacharyRizer: done

upload-reports/main.go
ZacharyRizer: Removed.

upload-reports/utils/config.go
ZacharyRizer: We had discussed leaving out this package for simplicity's sake before. Do you want me to add it and switch how the env vars are loaded?

upload-reports/utils/config.go
ZacharyRizer: So the target ENV is being used by the github actions to determine which github env vars to look at. And it isn't an input field, it is a multi select for the 4 standard environments.

upload-reports/utils/config.go
ZacharyRizer: Done

upload-reports/main.go
ZacharyRizer: Done

upload-reports/main.go
ZacharyRizer: I thought the goal for this was only for the github runner? It should just overwrite if something is there, but since this output csv only exists for the action runner, I don't think we need to.

upload-reports/utils/config.go
ZacharyRizer: Added the package with a conditional so it will work with or without.

cfarmer-fearless (approved):

PR #527, opened 2024-10-09
Simplify unmarshalling

whytheplatypus:
We may be able to go a bit further here, allowing us to remove the registration of targets step


emilio32-lang (approved):
reviewed

PR #528, opened 2024-10-09
UPLOAD-1798/gha-playwright

thetif:
- created playwright docker-compose file
- added playwright tests to `tus-upload-server-ci.yml`, testing all upload locations
- changed some playwright configuration
- removed the endpoint field from the upload form because it is unnecessary and possibly destructive
- updated playwright tests to remove endpoint field and make them a little more fault tolerant


tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -124,40 +121,34 @@ test.describe("Upload Status Page", () => {
...
 
         const fileChooser = await fileChooserPromise;
-        await fileChooser.setFiles('../upload-files/10KB-test-file');
+        await fileChooser.setFiles('./test-data/10KB-test-file');
 
         await expect((await uploadPatchResponsePromise).ok()).toBeTruthy()
         await expect((await uploadHeadResponsePromise).ok()).toBeTruthy()
 
         await page.reload();
+        // var refreshes = 0;
```
thetif, line 130: temporarily commented these tests out while we work out the new delivery situation, will probably need to make tests read from the delivery configs to know what the location should be

tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -124,40 +121,34 @@ test.describe("Upload Status Page", () => {
...
-        const fileDeliveryEicrContainer = page.locator('.file-delivery-container').nth(2)
-        await expect(fileDeliveryEicrContainer.getByRole('heading', { level: 2 })).toHaveText('EICR')
-        await expect(fileDeliveryEicrContainer.getByRole('heading', { level: 3 })).toHaveText('Delivery Status: SUCCESS')
-        await expect(fileDeliveryEicrContainer).toContainText(`Location: uploads/eicr/${uploadId}`)
-
-        const fileDeliveryNcirdContainer = page.locator('.file-delivery-container').nth(3)
-        await expect(fileDeliveryNcirdContainer.getByRole('heading', { level: 2 })).toHaveText('NCIRD')
-        await expect(fileDeliveryNcirdContainer.getByRole('heading', { level: 3 })).toHaveText('Delivery Status: SUCCESS')
-        await expect(fileDeliveryNcirdContainer).toContainText(`Location: uploads/ncird/${uploadId}`)
+        // const fileDeliveriesContainer = page.locator('.file-deliveries-container');
```
thetif, line 141: temporarily commented these tests out while we work out the new delivery situation, will probably need to make tests read from the delivery configs to know what the location should be

tests/smoke/playwright/test/upload-test.spec.ts
```diff
@@ -16,7 +16,7 @@ test.describe.skip('File Upload and Trace Response Flow', () => {
   const password = process.env.SAMS_PASSWORD;
   const url = process.env.UPLOAD_URL;
   const PS_API_URL = process.env.PS_API_URL;
-  const fileName = path.resolve(__dirname, '..', '..', 'upload-files', '10KB-test-file');
+  const fileName = path.resolve(__dirname, 'test-data', '10KB-test-file');
```
thetif, line 19: moved the test file into the same directory so it's all contained

upload-server/internal/appconfig/appconfig.go
```diff
@@ -34,58 +34,64 @@ type RootResp struct {
...
-	Metrics               MetricsConfig `env:", prefix=METRICS_"`
 
 	// TUSD
 	TusUploadPrefix string `env:"TUS_UPLOAD_PREFIX, default=tus-prefix"`
 
-	// UI
-	UIPort    string `env:"UI_PORT, default=:8081"`
+	// User Interface Configs
+	UIPort    string `env:"UI_PORT, default=8081"`
+	UIUrl     string
```
thetif, line 61: Before the port for the UI had to include the leading `:` which wasn't super intuitive, so I created the UIUrl field so that the port could just be the port

upload-server/internal/appconfig/appconfig.go
```diff
@@ -213,6 +219,12 @@ func ParseConfig(ctx context.Context) (AppConfig, error) {
 	ac.ServerFileEndpointUrl = ac.ServerUrl + ac.TusdHandlerBasePath
 	ac.ServerInfoEndpointUrl = ac.ServerUrl + ac.TusdHandlerInfoPath
 
+	if ac.UIPort != "" {
```
thetif, line 222: creating the UIUrl if the port is set, otherwise UIUrl is set to empty string so that the UI won't start up if the UI_PORT is nil

upload-server/internal/ui/ui.go
```diff
@@ -240,18 +241,19 @@ func GetRouter(uploadUrl string, infoUrl string) *mux.Router {
 			return
 		}
 
-		uploadUrl, err := url.JoinPath(uploadUrl, id)
+		uploadDestinationUrl, err := url.JoinPath(uploadUrl, id)
```
thetif, line 244: Removed the Endpoint field from the upload form because it wasn't needed, now passing in the uploadUrl that we are using as the endpoint for tus

tests/smoke/playwright/test/upload-test-pages.spec.ts
cfarmer-fearless: Would it be simpler to wait a certain amount of time (like 5 seconds) after an upload completes, and then check the status page as apposed to refreshing the page?  It's more of a "did we deliver the file in time" kind of approach, and that's what the kotlin tests do.

tests/smoke/playwright/test/upload-test-pages.spec.ts
```diff
@@ -124,40 +121,34 @@ test.describe("Upload Status Page", () => {
...
+        // while (await page.locator('.file-delivery-container').count() < targets.length && refreshes < 3) {
+        //     await page.reload();
+        //     refreshes++
+        // }
 
         const fileHeaderContainer= page.locator('.file-header-container')
         await expect(fileHeaderContainer.getByRole('heading', { level: 1 }).nth(0)).toHaveText(expectedFileName)
         await expect(fileHeaderContainer.getByRole('heading', { level: 1 }).nth(1)).toHaveText("Upload Status: Complete")
-        await expect(fileHeaderContainer).toContainText(`ID: ${uploadId}`)
+        await expect(fileHeaderContainer).toContainText(`ID: ${uploadId.split('+')[0]}`)
```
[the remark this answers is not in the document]
cfarmer-fearless: Hoping we can avoid this.  Ideally, these tests should have no awareness of the storage impl that the service under test is using.  Let's have a convo on how we are manipulating the ID at the presentation layer.

tests/smoke/playwright/test/upload-test-pages.spec.ts
cfarmer-fearless: Can the expected destinations be part of the test case? (again, this is how the kotlin tests do it)

tests/smoke/playwright/wait-for-it.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
cfarmer-fearless, line 1: Why is this needed?  Doesn't playwright take care of this already?

upload-server/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN mkdir -p ./configs/local
 COPY --from=builder /app/configs/local/deliver.yml ./configs/local/deliver.yml
 
 # Create the 'app' user and group
-RUN addgroup -S app && adduser -S -G app app
+RUN addgroup -S -g 101 app && adduser -S -u 100 -G app app
```
cfarmer-fearless, line 32: What's the purpose of this change?  Maybe belongs in another PR

tests/smoke/playwright/test/upload-test-pages.spec.ts
cfarmer-fearless: Regarding the new delivery config, you should be able to use the `DEX_DELIVERY_CONFIG_FILE` env var and the yml files defined in upload-server/configs/testing

upload-server/internal/appconfig/appconfig.go
cfarmer-fearless: If this is the only reason for splitting this config field, maybe we can change the underlying code to just include the `:`.  Unless there's a need for the service to be aware of the hostname, which I don't see any need for.

cfarmer-fearless (changes requested):

upload-server/Dockerfile
thetif: when Candace made her change to the previous part, it didn't seem to want to work for me until I did that. I can back it out and make it's not necessary now

tests/smoke/playwright/wait-for-it.sh
thetif: Playwright has a solution where it will start the server for you, but we can't use it because of the way the project is set up. If the server were Javascript it would be easy to do it. This is a work around because sometimes it can take much longer for the upload-server to start than the playwright, we can extend the timeout for the first test in every suite but that's not very durable.

tests/smoke/playwright/test/upload-test-pages.spec.ts
thetif: yeah we definitely need to switch to using the config file, but that was outside the scope of this ticket. I had made that fix before the delivery stuff changed and then commented it out when it did.

thetif: we need to make a separate ticket to update everything based on the new configs, should make things easier too

tests/smoke/playwright/test/upload-test-pages.spec.ts
thetif: yes, that's what I was trying to do before the delivery stuff changed, and the delivery configs make it every easier to dynamically check for destinations

upload-server/internal/appconfig/appconfig.go
thetif: the NewServer expects it to be in that format and I was just going to set it there, but it seemed more sticking with the convention to set it in the appconfig because that's where we are creating all the other configs and maybe we will want to give the UI a hostname eventually

tests/smoke/playwright/package.json
```diff
@@ -5,7 +5,9 @@
   "main": "dist/main.js",
   "scripts": {
     "build": "npx tsc",
-    "test": "npx playwright test"
+    "wait": "./wait-for-it.sh ${UI_URL:-http://localhost:8081}",
+    "test": "npm run wait; npx playwright test",
```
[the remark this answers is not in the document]
thetif: Yes I was thinking we didn't really need it, but I thought it probably didn't hurt to have it check before running. It doesn't only wait for localhost:8081, it waits for whatever UI_URL is set to and defaults to localhost:8081 if it's not set.

tests/smoke/playwright/playwright.config.ts
```diff
@@ -1,19 +1,49 @@
...
+  workers: process.env.CI ? 1 : 4,
+
+  // Fail the build on CI if you accidentally left test.only in the source code.
+  forbidOnly: !!process.env.CI,
 
-  // Configure retries for flaky tests
-  retries: 0,
+  // If a test fails, retry it additional 2 times
+  // Retry on CI only.
+  retries: process.env.CI ? 2 : 0,
```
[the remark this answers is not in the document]
thetif: we can set it 0 all the time

tests/smoke/playwright/test/upload-test-accessibility.spec.ts
```diff
@@ -31,7 +31,7 @@ test.describe('Upload Manifest Page', () => {
 
 test.describe('File Upload Page', () => {
     test(`Checks accessibliity for the upload page for the dextesting/testevent1 manifest`, async ({ page }) => {
-        await page.goto(`/manifest?data_stream_id=dextesting&data_stream_route=testevent1`);
+        await page.goto(`/manifest?data_stream_id=dextesting&data_stream_route=testevent1`, { waitUntil: 'load' });
```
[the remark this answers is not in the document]
thetif: I think we could probably get rid of those, that was my first attempt before adding the wait-for-it

.github/workflows/e2e-test-template.yml
```diff
@@ -0,0 +1,40 @@
...
+    env:
+      CI: true
+      AZURITE_STORAGE_KEY: ${{ (inputs.STORAGE_TYPE == 'azure') && secrets.AZURITE_STORAGE_KEY }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install podman compose
+        run: pip3 install podman-compose
+      - name: Run E2E Tests
+        id: test
+        run: podman-compose -f ${{ inputs.COMPOSE_FILENAME }} up --build --exit-code-from playwright --abort-on-container-exit
```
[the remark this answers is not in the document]
thetif: oh good point, I didn't bring the continue-on-error over

tests/smoke/playwright/test/upload-test-pages.spec.ts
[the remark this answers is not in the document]
thetif: Were we not able so resolve that filename tus string issue in S3? Or did Nicole run out of time before leaving? We should make a ticket to fix it if she just wasn't able to get to it in time

.github/workflows/e2e-test-template.yml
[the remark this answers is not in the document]
cyber-decker: Perfect. I think for now, at least for the tests, if the playwright tests fail, lets be notified, but not stop the presses.   I want to make sure we can really trust the tests before we have them telling us that it should stop.

cfarmer-fearless (approved):

PR #529, opened 2024-10-09
Fix for shouldTranslateMetadataGivenV1SenderManifest tests

GSA2024:
The code fixes the 404 issues by constructing the expected filename accurately and checking metadata mappings to ensure the key -value pair match as expected.


cfarmer-fearless:
Closing in favor of #532

PR #531, opened 2024-10-11
add covid all monthly

cfarmer-fearless:


AJKimpo (changes requested):
adding `covid-all-monthly-vaccination` looks good

see line 84: change data stream id to `rsv-prevention`

AJKimpo (approved):
looks good

PR #532, opened 2024-10-11
Upload 1747/cloud agnostic e2e tests

cfarmer-fearless:
Making the end to end kotlin TestNG tests more cloud agnostic and not rely on cloud services to validate results.  The primary change here is rather than going in to Azure to validate results or collect information to validate against, we will use the /info endpoint of the upload api to validate results against and trust those results there.

Changes:

- Add an environment variable to the local.properties files to name the environment being targeted. 
- Remove connecting to the Azure Blob Service from tests and instead utilize the existing uploadClient to get details from the '/info' endpoint given a UID
- change tests to read from a "test case" with manifest details and other validations rather than just a manifest
- Change assertions to validate from the '/info' endpoint results and make assertions more descriptive and informative when they fail.
- Include details to assert against the target path file location and provide details in the test case about the format of this path using tokens that can be replaced with collected data
- Simplify V1 manifest tests to pass (but these will eventually go away as V1 is completely phased out)
- Update Processing Status tests to use a test case rather than a manifest

### DIFF
--- a/tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
+++ b/tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
@@ -1,5 +1,4 @@
 import com.azure.identity.ClientSecretCredentialBuilder
-import com.azure.storage.blob.BlobClient
 import dex.DexUploadClient
 import org.testng.Assert
 import org.testng.ITestContext
@@ -16,7 +15,7 @@ import kotlin.collections.HashMap
 @Test()
 class FileCopy {
     private val testFile = TestFile.getResourceFile("10KB-test-file")
-    private val authClient = DexUploadClient(EnvConfig.UPLOAD_URL)
+    private val dexUploadClient = DexUploadClient(EnvConfig.UPLOAD_URL)
     private val dexBlobClient = Azure.getBlobServiceClient(EnvConfig.DEX_STORAGE_CONNECTION_STRING)
     private val edavBlobClient = Azure.getBlobServiceClient(
         EnvConfig.EDAV_STORAGE_ACCOUNT_NAME,
@@ -37,7 +36,7 @@ class FileCopy {
 
     @BeforeTest(groups = [Constants.Groups.FILE_COPY])
     fun beforeFileCopy() {
-        authToken = authClient.getToken(EnvConfig.SAMS_USERNAME, EnvConfig.SAMS_PASSWORD)
+        authToken = dexUploadClient.getToken(EnvConfig.SAMS_USERNAME, EnvConfig.SAMS_PASSWORD)
     }
 
     @BeforeMethod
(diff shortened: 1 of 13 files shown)


tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
```diff
@@ -51,41 +39,47 @@ class FileCopy {
...
-            expectedBlobClient = routingContainerClient.getBlobClient(expectedFilename)
-
-            Assert.assertNotNull(expectedBlobClient)
-            Assert.assertEquals(expectedBlobClient!!.properties.blobSize, testFile.length())
+        Thread.sleep(5000)
+        val uploadInfo = dexUploadClient.getFileInfo(uid, authToken)
+
+        // Check File Info
+        val expectedBytes = "10240"
+        Assert.assertEquals(uploadInfo.fileInfo.sizeBytes.toString(), expectedBytes)
```
cfarmer-fearless, line 51: Can keep the expected bytes as an int

tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
```diff
@@ -51,41 +39,47 @@ class FileCopy {
...
+
+        val expectedDeliveryNames = case.deliveryTargets?.map{ it.name }?.sorted()
+        val actualDeliveryNames = uploadInfo.deliveries?.map{ it.name }?.sorted()
+        Assert.assertEquals(actualDeliveryNames, expectedDeliveryNames, "Actual delivery targets do not match expected targets")
+
+    
+        val currentDateTime = ZonedDateTime.now(TimeZone.getTimeZone("GMT").toZoneId())
+        
+        uploadInfo.deliveries?.forEach { delivery ->
+            Assert.assertEquals(delivery.status, "SUCCESS") // remove the assertion above?
```
cfarmer-fearless, line 68: I agree with this comment.  `uploadInfo.deliveries.all` will loop over the collection.  If we have to loop over it anyway to check the path, just do it here.

tests/smoke/kotlin/src/test/kotlin/ProcStat.kt
```diff
@@ -158,16 +158,18 @@ class ProcStat {
                     assertNotNull(destinationName, "Destination name is missing for $schemaName")
                     log.info("Blob File Copy - Source URL: $sourceUrl, Destination URL: $destinationUrl")
 
-                    val expectedDestinations = config.copyConfig.targets
+//                    val expectedDestinations = config.copyConfig.targets
 
-                    val actualDestination = expectedDestinations.any { expectedDest ->
-                        destinationName?.trim()?.equals(expectedDest.trim(), ignoreCase = true) ?: false
-                    }
+                    val expectedDestinations = case.deliveryTargets!!.map{ it.name }.sorted()
```
cfarmer-fearless, line 163: Prefer `?` over `!!` operator and just skip the assert if the case has no targets or is null.

tests/smoke/kotlin/src/test/kotlin/util/DataProvider.kt
```diff
@@ -83,18 +84,28 @@ class DataProvider {
...
+}
+
+data class TestCase(
+    @JsonProperty("manifest")val manifest: Map<String, String>,
+    @JsonProperty("delivery_targets") val deliveryTargets: List<Target>?
+)
+
+data class Target(
+    @JsonProperty("name") val name: String,
+    @JsonProperty("path_template") val pathTemplate: Map<String, String>
```
cfarmer-fearless, line 110: Create a nested data class and assign it to the `pathTemplate` type instead of a generic `Map<String, String>`.  That way, you'll be able to enforce that every path template needs a value for every environment you want to target because you'll get a runtime error if you don't provide all the values and try to deserialize.

```kotlin
data class PathTemplateByEnvironment(val local: String, val dev: String, val test: String, val stg: String)
```

Be sure to use `@JsonProperty` as well since the key names in the json are capital case.

tests/smoke/kotlin/src/test/kotlin/util/EnvConfig.kt
```diff
@@ -9,6 +9,7 @@ class EnvConfig {
         private val properties = if (propFile.exists()) Properties().apply {
             load(File("local.properties").inputStream())
         } else null
+        val ENVIRONMENT: String = properties?.getProperty("environment") ?: System.getenv("ENVIRONMENT")
```
cfarmer-fearless, line 12: Define an enum class for this and convert the string value to it.  Something like
```kotlin
enum class Environment {
  LOCAL("LOCAL"), DEV("DEV"), TEST("TEST"), STG("STG")
}
```

That way the tests will fail right away if this environment variable isn't set to something you expect.  In addition, it'll prevent a downstream key error if you try to do `pathTemplate.get(ENVIRONMENT)`

cfarmer-fearless (commented):
Looks great overall!  Just a few minor comments

tests/smoke/kotlin/src/test/kotlin/util/DataProvider.kt
cyber-decker: I think this might need a better solution.  Pulling it in like this makes the pathTemplate for each environment difficult to access when we want to just get it for the current environment unless maybe we use reflection?  With the map, at least we could access it by the environment variable being set.  Might need some help with figuring this one out.

whytheplatypus (approved):

PR #533, opened 2024-10-15
deliminates data stream and route with an underscore

cfarmer-fearless:
This patch addresses an issue with how we use filenames to define data streams and routes for manifest validation.  Currently, data streams and routes provided in a sender manifest are used to read in the manifest config json file by joining the data stream and route strings with a "-" (dash) character.  This made it difficult to determine the datastream and route given the filename since some data streams had dashes within their values.  This allowed for unintended behavior in the UI by allowing some invalid inputs to be deemed valid.

For example, if we wanted to pick the datastream "covid-all-monthly-vaccination" and the route of "csv", we could say the datastream is actually "covid-all" and the route is "monthly-vaccination-csv".  The service would simple concatinate these two values with a dash, and would be able to resolve the json file.  Obviously, this input combination is invalid and the user should instead see a "not found" message.

To fix this, I've renamed all of the V2 config files to use an "_" (underscore) delimiter between the datastream and route instead of a dash.  That way, it's more clear where the two parts of information are, and gives the desired behavior in the UI.


upload-server/testing/cli_test.go
```diff
@@ -289,7 +289,7 @@ func TestGetFileDeliveryPrefixDate(t *testing.T) {
 		"data_stream_id":    "test_stream",
 		"data_stream_route": "test_route",
 	}
-	metadata.Cache.SetConfig("v2/test_stream-test_route.json", &validation.ManifestConfig{
+	metadata.Cache.SetConfig("v2/test_stream_test_route.json", &validation.ManifestConfig{
```
whytheplatypus, line 292: to stick with this new convention update this to match.

```suggestion
	metadata.Cache.SetConfig("v2/test-stream_test-route.json", &validation.ManifestConfig{
```

whytheplatypus (changes requested):

thetif (approved):
LGTM

upload-server/testing/cli_test.go
cfarmer-fearless: Done

whytheplatypus (approved):

PR #535, opened 2024-10-18
Decouple Delivery Targets From Route Groups

cfarmer-fearless:
This is an iteration on the deliver configuration schema.  The goal is to decouple delivery targets from routing groups.  This will improve robustness by preventing possible misconfiguration of creating multiple delivery targets with the same name.  With this, we want to make sure the server fails to start if the deliver yml is misconfiguration in this way.

### DIFF
--- a/upload-server/configs/testing/s3_destinations.yml
+++ b/upload-server/configs/testing/s3_destinations.yml
@@ -1,102 +1,76 @@
 version: 1.0
 
-s3-edav: &s3-edav
-  bucket_name: edav
-  access_key_id: minioadmin
-  secret_access_key: minioadmin
-  endpoint: "http://minio:8000"
-  region: us-east-1
+targets:
+  - name: edav
+    type: s3
+    bucket_name: edav
+    access_key_id: minioadmin
+    secret_access_key: minioadmin
+    endpoint: "http://minio:8000"
+    region: us-east-1
 
-s3-ehdi: &s3-ehdi
-  bucket_name: ehdi
-  access_key_id: minioadmin
-  secret_access_key: minioadmin
-  endpoint: "http://minio:8000"
-  region: us-east-1
+  - name: ehdi
+    type: s3
+    bucket_name: ehdi
+    access_key_id: minioadmin
+    secret_access_key: minioadmin
+    endpoint: "http://minio:8000"
+    region: us-east-1
 
-s3-eicr: &s3-eicr
-  bucket_name: eicr
-  access_key_id: minioadmin
-  secret_access_key: minioadmin
-  endpoint: "http://minio:8000"
-  region: us-east-1
+  - name: eicr
+    type: s3
+    bucket_name: eicr
+    access_key_id: minioadmin
+    secret_access_key: minioadmin
+    endpoint: "http://minio:8000"
+    region: us-east-1
 
-s3-ncird: &s3-ncird
-  bucket_name: ncird
-  access_key_id: minioadmin
-  secret_access_key: minioadmin
-  endpoint: "http://minio:8000"
-  region: us-east-1
-
-# if needed, properties of an anchor can be overridden
-# in a delivery target
-#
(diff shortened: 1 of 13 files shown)


upload-server/internal/postprocessing/postprocessor.go
```diff
@@ -54,7 +54,7 @@ func ProcessFileReadyEvent(ctx context.Context, e *event.FileReady) error {
 	}
 	dataStreamId := metadata.GetDataStreamID(e.Metadata)
 	dataStreamRoute := metadata.GetDataStreamRoute(e.Metadata)
-	d, ok := delivery.GetDestinationTarget(dataStreamId, dataStreamRoute, e.DestinationTarget)
+	d, ok := delivery.GetGroupTarget(dataStreamId, dataStreamRoute, e.DestinationTarget)
```
whytheplatypus, line 57: any concerns with reducing this down to `delivery.GetTarget(e.DestinationTarget)`?

cfarmer-fearless: Yes.  It's a many:many relationship between groups and targets, so in order to get a target, we need to know the data stream and route, as well as the target name

upload-server/configs/testing/s3_destinations.yml
```diff
@@ -1,102 +1,76 @@
 version: 1.0
 
-s3-edav: &s3-edav
-  bucket_name: edav
-  access_key_id: minioadmin
-  secret_access_key: minioadmin
-  endpoint: "http://minio:8000"
-  region: us-east-1
+targets:
+  - name: edav
```
whytheplatypus, line 4: this can be turned into a map so it deserializes naturally into a `map[string]destination`

upload-server/configs/testing/s3_destinations.yml
```diff
@@ -1,102 +1,76 @@
...
+    bucket_name: ncird
+    access_key_id: minioadmin
+    secret_access_key: minioadmin
+    endpoint: "http://minio:8000"
+    region: us-east-1
+routing_groups:
   - data_stream_id: pulsenet
     data_stream_route: localsequencefile
     delivery_targets:
       - name: edav
```
whytheplatypus, line 39: the `name` key is redundant this can be an array of strings (similar to something like `depends_on` on in a compose file

upload-server/configs/testing/s3_destinations.yml
```diff
@@ -1,102 +1,76 @@
...
       - name: ehdi
-        type: s3
         path_template: $EHDI_PATH_TEMPLATE
-        <<: *s3-ehdi
   - data_stream_id: eicr
     data_stream_route: fhir
     delivery_targets:
       - name: eicr
-        type: s3
         path_template: $EICR_PATH_TEMPLATE
```
whytheplatypus, line 49: move this into the target definition.

upload-server/internal/delivery/deliver.go
```diff
@@ -209,10 +221,9 @@ func RegisterAllSourcesAndDestinations(ctx context.Context, appConfig appconfig.
 			Prefix:     appConfig.TusUploadPrefix,
 		}
 	}
-	RegisterSource("upload", src)
+	RegisterSource(UploadSrc, src)
 
-	targets = append(targets, src)
-	if err := health.Register(targets...); err != nil {
+	if err := health.Register(getDeliveryHealthChecks([]Source{src}, dests)...); err != nil {
```
whytheplatypus, line 226: call this twice rather than use getDeliveryHealthChecks for clarity

whytheplatypus (changes requested):

upload-server/configs/testing/s3_destinations.yml
cfarmer-fearless: The path template? That should be decided by the route group, right?

whytheplatypus (approved):

PR #538, opened 2024-10-21
#UPLOAD-1826 update all date formats to RFC3339Nano

marcia-schulman:
Update all time formats to include nanosecond precision.


whytheplatypus (approved):
yup looks good

PR #541, opened 2024-10-23
delete raw uploads script

rohitpanwar:


upload-scripts/delete-raw-uploads.go
```diff
@@ -0,0 +1,184 @@
...
+
+func deleteBlobs(serviceClient *azblob.Client) {
+
+	for _, blobName := range blobsToDelete {
+
+    	log.Printf("Deleting Blob - %s", blobName)
+    	_, err := serviceClient.DeleteBlob(context.Background(), CONTAINER, blobName, nil)
+    	if err != nil {
+        	log.Printf("Failed to delete blob: %v", err)
+    	}
```
cfarmer-fearless, line 182: Fix formatting here

cfarmer-fearless: Run `go fmt ./...`

cfarmer-fearless (changes requested):
Some improvements to be made.  Considering this needs to search through millions of files, I think it's worth it to spend the development time now to make this more robust so the searching is faster down the road.  Specifically with parallelism with go routines and blob indexing.

I also suggest renaming the "upload-scripts" folder to "upload-data-gov" as this tool might turn into a more general data governance tool.

cfarmer-fearless:
Once this gets more solidified, please add a basic readme

upload-scripts/main.go
```diff
@@ -0,0 +1,258 @@
...
+			log.Printf("Blob content is not JSON: %s\nContent:\n%s\n", blobName, string(body))
+			return
+
+		} else {
+
+			metadata, metadataExists := jsonData["MetaData"].(map[string]interface{})
+
+			if metadataExists {
+			
+				data_stream :=  metadata["meta_destination_id"]
```
cfarmer-fearless, line 201: Will this skip files with `data_stream_id` and `data_stream_route` in their metadata?

upload-scripts/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if *DATA_STREAM == "" {
+		log.Fatalf("DATA_STREAM is not set")
+	}
+
+	if *ROUTE == "" {
+		log.Fatalf("ROUTE is not set")
+	}
+
+	if(ACCOUNT_KEY == "") {
```
cfarmer-fearless, line 62: This will always evaluate to true

upload-scripts/main.go
```diff
@@ -0,0 +1,258 @@
...
+		// IF not set for local runs pull from the environment
+		envKey := fmt.Sprintf("ACCOUNT_KEY_%s", strings.ToUpper(*TARGET_ENV))
```
cfarmer-fearless, line 64: So the actual environment variable that users need to set is `ACCOUNT_KEY_DEV` or `ACCOUNT_KEY_TST` etc?  And not just `ACCOUNT_KEY` as the readme suggests?  Kinda confusing and seems unnecessary.  Should just be `ACCOUNT_KEY`

cfarmer-fearless (changes requested):
See above comments.  Also got the following error when running on GFE
```
2024/11/05 14:51:58 Failed to list blobs: GET https://ocioededataexchangedev.blob.core.windows.net/bulkuploads
--------------------------------------------------------------------------------
RESPONSE 403: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
ERROR CODE: AuthenticationFailed
--------------------------------------------------------------------------------
﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:68615e98-001e-0059-03bc-2fe995000000
Time:2024-11-05T19:51:58.8434953Z</Message><AuthenticationErrorDetail>The MAC signature found in the HTTP request 'oRfmwqg6nJGDEGPlJQRbeE3ed+6TNGSsbNvNW7r0/LE=' is not the same as any computed signature. Server used following string to sign: 'GET











x-ms-date:Tue, 05 Nov 2024 19:51:58 GMT
x-ms-version:2024-08-04
/ocioededataexchangedev/bulkuploads
comp:list
prefix:tus-prefix/
restype:container'.</AuthenticationErrorDetail></Error>
--------------------------------------------------------------------------------
exit status 1
```

upload-scripts/main.go
```diff
@@ -0,0 +1,422 @@
...
+		slog.Error("error fetching filtered blobs", "error", err)
+		return
+	}
+
+	for _, res := range rsp.Blobs {
+		totalSearched++
+		uid := *res.Name
+		if strings.Contains(*res.Name, ".info") {
+			uid = strings.Split(*res.Name, ".")[0]
+		}
```
whytheplatypus, line 205: if both match as tags this will return a duplicate result

upload-scripts/main.go
```diff
@@ -0,0 +1,422 @@
...
+	return out
+}
+
+func worker(ctx context.Context, c <-chan pageItemResult, o chan<- matchResult, containerClient *container.Client, criteria map[string]string) {
+	for r := range c {
+		uid := *r.item.Name
+
+		if strings.Contains(*r.item.Name, ".info") {
+			uid = strings.Split(*r.item.Name, ".")[0]
+		}
```
whytheplatypus, line 231: will result in duplicate results. could skip if it's an `info` file?

whytheplatypus (changes requested):

upload-scripts/main.go
cfarmer-fearless: Good point.  I think you're right.  I need to flip the logic here.  Skip if .info; otherwise start trying to match.  And if we are inspecting an actual file, we can pretty much guarantee the info file exists.

cfarmer-fearless: Problem with this though is it may leave some zombie .info files.  Maybe that's ok

whytheplatypus (approved):

PR #542, opened 2024-10-24
set delivery name on builder init

cfarmer-fearless:
This patch fixes an issue where delivery info for a blob copy report won't get set if there is a delivery failure.  The fix is to simply set the info when the builder for the report is instantiated.

### DIFF
--- a/upload-server/internal/postprocessing/postprocessor.go
+++ b/upload-server/internal/postprocessing/postprocessor.go
@@ -35,7 +35,14 @@ func ProcessFileReadyEvent(ctx context.Context, e *event.FileReady) error {
 		"1.0.0",
 		reports.StageFileCopy,
 		e.UploadId,
-		reports.DispositionTypeAdd).SetStartTime(time.Now().UTC())
+		reports.DispositionTypeAdd).SetStartTime(time.Now().UTC()).SetContent(reports.FileCopyContent{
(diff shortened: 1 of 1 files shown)


whytheplatypus (changes requested):
the bit setting this info at the end of the function can be removed

cfarmer-fearless:
> the bit setting this info at the end of the function can be removed

I believe we still need it so we can set the destination URL, which we only have after successful delivery.  And in the current impl, the content is immutable, so can only set it to a new instance with the setter method

cyber-decker:
Functionally this looks good.   I gave this a test locally and the info endpoint seems to be adding the name as we expect now.  Here's the comparision of two outputs that ran locally.

[fixed-code-2fcd10dd-df43-4aa3-a60c-8f04493938f1.json](https://github.com/user-attachments/files/17524592/fixed-code-2fcd10dd-df43-4aa3-a60c-8f04493938f1.json)
[current-code-in-main-c13953da-6b25-403a-9a23-92ea809f045a.json](https://github.com/user-attachments/files/17524602/current-code-in-main-c13953da-6b25-403a-9a23-92ea809f045a.json)

I was a little weirded out by all the delivery entries we are seeing but that seems to be a result of something else and is that way currently in main.

cyber-decker (approved):
Looks good, seems to resolve the issue.

PR #550, opened 2024-10-29
Integrate code from lib-cloud-proxy-go into routing delivery

marcia-schulman:
Adding functionality to Source and Destination types to support cross-cloud large file copies.


upload-server/internal/delivery/azure.go
```diff
@@ -13,11 +20,23 @@ import (
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/storeaz"
 )
 
+const maxPartsAzure = 50000
```
cfarmer-fearless, line 23: Wonder if these two should be at the config level.  They look like things that belong in a helm chart.

upload-server/internal/delivery/azure.go
```diff
@@ -96,23 +137,145 @@ func (ad *AzureDestination) Client() (*container.Client, error) {
...
+				select {
+				case errCh <- err:
+					// error was set
+				default:
+					// some other error is already set
+				}
+				cancel()
+			}
+		}(id)
+		if routines >= concurrency {
```
cfarmer-fearless, line 238: This check and additional wait seem unnecessary.  The wait on line 243 should suffice

upload-server/internal/delivery/azure.go
```diff
@@ -96,23 +137,145 @@ func (ad *AzureDestination) Client() (*container.Client, error) {
...
+			wg.Wait()
+			routines = 0
+		}
+	}
+	wg.Wait()
+	select {
+	case err := <-errCh:
```
cfarmer-fearless, line 245: What's the behavior we want here?  Do we need to make sure all blocks staged successfully?  Is it all or nothing?  Or can we recover if we get errors for only a few?

upload-server/internal/delivery/deliver.go
```diff
@@ -26,13 +26,31 @@ import (
...
+	Reader(context.Context, string) (io.Reader, error)
+	GetMetadata(context.Context, string) (map[string]string, error)
+	GetSignedObjectURL(ctx context.Context, containerName string, objectPath string) (string, error)
+	GetSourceFilePath(string) string
+	SourceType() string
+	Container() string
+}
+
+type Destination interface {
+	Copy(ctx context.Context, path string, source *Source, metadata map[string]string, length int64, concurrency int) (string, error)
```
cfarmer-fearless, line 50: Do you need to pass in path, metadata, and length if passing in source?  Can those things be gotten from the source?

upload-server/internal/delivery/azure.go
marcia-schulman: changed to use work pool

upload-server/internal/delivery/azure.go
marcia-schulman: due to automatic staging retries, if it gets this far, it is best to abort.

upload-server/internal/delivery/deliver.go
marcia-schulman: no :)

upload-server/internal/appconfig/appconfig.go
```diff
@@ -84,6 +84,8 @@ type AppConfig struct {
 	LocalReportsFolder string `env:"LOCAL_REPORTS_FOLDER, default=./uploads/reports"`
 	LocalEventsFolder  string `env:"LOCAL_EVENTS_FOLDER, default=./uploads/events"`
 
+	// Cloud Transfer Configs
+	MaxConcurrency int `env:"MAX_CONCURRENCY, default=500"`
```
cfarmer-fearless, line 88: This can be named a bit more specific like `MAX_DELIVERY_CONCURRENCY` since we have another environment variable for setting routing event handler worker pools.

upload-server/internal/delivery/azure.go
```diff
@@ -13,11 +20,23 @@ import (
 	"github.com/cdcgov/data-exchange-upload/upload-server/internal/storeaz"
 )
 
+const maxPartsAzure = 50000 // maximum number of parts per block blob in Azure Storage
```
cfarmer-fearless, line 23: Let's include the units in the names, and also move this to app config

upload-server/internal/delivery/azure.go
```diff
@@ -38,7 +57,29 @@ func (ad *AzureSource) GetMetadata(ctx context.Context, tuid string) (map[string
 	if err != nil {
 		return nil, err
 	}
-	return storeaz.DepointerizeMetadata(resp.Metadata), nil
+	props := storeaz.DepointerizeMetadata(resp.Metadata)
+	props["last_modified"] = resp.LastModified.Format(time.RFC3339Nano)
```
cfarmer-fearless, line 61: Should setting the last modified and content length properties be moved to a higher level operation like a hook?  Setting them within `GetMetadata` seems like a side-effect

upload-server/internal/delivery/azure.go
```diff
@@ -38,7 +57,29 @@ func (ad *AzureSource) GetMetadata(ctx context.Context, tuid string) (map[string
...
+	props["content_length"] = strconv.FormatInt(*resp.ContentLength, 10)
+	return props, nil
+}
+
+func (ad *AzureSource) GetSignedObjectURL(_ context.Context, _, objectPath string) (string, error) {
+	path := ad.GetSourceFilePath(objectPath)
+	sourceBlob := ad.FromContainerClient.NewBlockBlobClient(path)
+	sourceURL, er := sourceBlob.GetSASURL(sas.BlobPermissions{
+		Read:   true,
```
cfarmer-fearless, line 70: Might be best if these permissions are passed into the function, that way we can do least privilege if need be.

upload-server/internal/delivery/azure.go
```diff
@@ -97,17 +143,151 @@ func (ad *AzureDestination) Client() (*container.Client, error) {
...
+	client, e := ad.Client()
+	if e != nil {
+		return "", fmt.Errorf("unable to obtain Azure container client: %v", e)
+	}
+	var partSize int64 = size5MB
+	if length > size5MB*maxPartsAzure {
+		// we need to increase the Part size
+		partSize = length / maxPartsAzure
+	}
+	numChunks := length / partSize
```
cfarmer-fearless, line 196: Do we need to worry about any truncation here?

upload-server/internal/delivery/azure.go
```diff
@@ -97,17 +143,151 @@ func (ad *AzureDestination) Client() (*container.Client, error) {
...
+	defer wg.Done()
+	for job := range jobsCh {
+		_, err := blockBlobClient.StageBlockFromURL(ctx, job.BlockId, sourceSignedURL,
+			&blockblob.StageBlockFromURLOptions{
+				Range: job.Range,
+			})
+
+		if err != nil {
+			select {
+			case errCh <- err:
```
cfarmer-fearless, line 262: Do we need this select here?  Don't we just want to do `errCh <- err` and return?

upload-server/internal/delivery/deliver.go
```diff
@@ -221,25 +235,26 @@ func RegisterAllSourcesAndDestinations(ctx context.Context, appConfig appconfig.
...
-	if err != nil {
-		return "", err
+	length, e := strconv.ParseInt(manifest["content_length"], 10, 64)
+	if e != nil {
+		length = 1
 	}
-	if rc, ok := r.(io.Closer); ok {
-		defer rc.Close()
+	maxConcurrency := appconfig.LoadedConfig.MaxConcurrency
+	if maxConcurrency <= 0 {
```
cfarmer-fearless, line 248: We can just default the `MaxConcurrency` to 500 in the app config

upload-server/internal/delivery/deliver.go
```diff
@@ -221,25 +235,26 @@ func RegisterAllSourcesAndDestinations(ctx context.Context, appConfig appconfig.
...
 	if err != nil {
 		return "", err
 	}
 
-	r, err := s.Reader(ctx, id)
-	if err != nil {
-		return "", err
+	length, e := strconv.ParseInt(manifest["content_length"], 10, 64)
+	if e != nil {
+		length = 1
```
cfarmer-fearless, line 245: Does setting the length to 1 mean we fall back to the old way of copying the file?

upload-server/internal/delivery/s3.go
```diff
@@ -123,6 +168,152 @@ func (sd *S3Destination) Client() *s3.Client {
...
+}
+
+func (sd *S3Destination) copyPartWorker(ctx context.Context, jobsCh <-chan s3.UploadPartCopyInput,
+	responseCh chan<- types.CompletedPart, errCh chan<- error, wg *sync.WaitGroup) {
+	defer wg.Done()
+	for job := range jobsCh {
+		uploadPartResp, err := sd.toClient.UploadPartCopy(ctx, &job)
+		if err != nil {
+			select {
+			case errCh <- err:
```
cfarmer-fearless, line 303: Is this select necessary?  Just `errCh <- err` and return, right?

cfarmer-fearless (changes requested):
Some relatively minor polishing but overall looks great.  I also want to see the unit tests a bit more robust by creating any storage containers on the fly so they can work in a CI pipeline

cfarmer-fearless:
Closing due to staleness

PR #551, opened 2024-10-31
PHDO-87: Use dex_ingest_datetime for file delivery

whytheplatypus:
before this the current time at delivery was used to construct the destination path. This changes that to use the time the file submission started from the sender.

### DIFF
--- a/upload-server/cmd/cli/route.go
+++ b/upload-server/cmd/cli/route.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"slices"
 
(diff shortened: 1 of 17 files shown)


cfarmer-fearless (changes requested):
Looks great!  Please remove the copy config struct from validation.go https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/internal/metadata/validation/validation.go#L21

Similarly, remove the copy config JSON from the upload config json files

cfarmer-fearless (approved):
Looks good!  I think adding the path to the event is a good move.  One less thing for the handler to worry about

PR #552, opened 2024-11-01
PHDO-48/remove version from manifest

cfarmer-fearless:
This change simplifies the manifest logic within the service by removing the handling of the specific version of the manifest.  There is no longer a "v1" or "v2".  There is only one manifest and one schema.  In addition, hydration of v1 manifests has also been removed.


AJKimpo (changes requested):
A few things missed:

## Remove v1 Kotlin file `valid_manifests_v1.json`
- https://github.com/CDCgov/data-exchange-upload/blob/PHDO-48/remove-manifest-v1/tests/smoke/kotlin/src/test/resources/valid_manifests_v1.json

## Update Kotlin file to remove v1 references `invalid_manifests_required_fields.json` - lines 2-13
- https://github.com/CDCgov/data-exchange-upload/blob/PHDO-48/remove-manifest-v1/tests/smoke/kotlin/src/test/resources/invalid_manifests_required_fields.json

## Updates to test config `dextesting_testevent1.json`
- remove all copy_config targets except for "edav"
- this config does not include lines for "allowed values" - should it include those, whether set to null or set to explicit values to align with other configs?
- https://github.com/CDCgov/data-exchange-upload/blob/PHDO-48/remove-manifest-v1/upload-configs/dextesting_testevent1.json

AJKimpo (approved):
lgtm

whytheplatypus (approved):

PR #553, opened 2024-11-01
hotfix/lowercase config path and check for compat config on delivery

cfarmer-fearless:


cfarmer-fearless:
Closing in favor of removing version 1 manifest

PR #563, opened 2024-11-05
run upload server as standalone subscriber

emilio32-lang:
default instance: configure publisher other than MemoryPublisher
standalone subscriber: configure subscriber to listen for events from publisher from the default instance


whytheplatypus (changes requested):
I'm guessing this pr is flipped? should be `standalone-mode` -> `main`

PR #566, opened 2024-11-06
Updating NCIRD targets for tests and handling + symbol

cyber-decker:
TestNG test config fixes

- Changing the NCIRD targets for the DEV environment
- Handling `+` symbol in the actual file locations so they don't get converted to spaces

### DIFF
--- a/tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
+++ b/tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
@@ -60,7 +60,8 @@ class FileCopy {
         val currentDateTime = ZonedDateTime.now(TimeZone.getTimeZone("GMT").toZoneId())
         uploadInfo.deliveries?.forEach { delivery ->
             Assert.assertEquals(delivery.status, "SUCCESS") // remove the assertion above?
-            val actualLocation = URLDecoder.decode(delivery.location, StandardCharsets.UTF_8.toString())
+            val encodedLocation = delivery.location.replace("+", "%2B")
+            val actualLocation = URLDecoder.decode(encodedLocation, StandardCharsets.UTF_8.toString())
             val pattern = case.deliveryTargets?.find{ it.name == delivery.name}?.pathTemplate?.get(EnvConfig.ENVIRONMENT)
             val expectedLocation = pattern
                 ?.replace("{dataStream}", case.manifest["data_stream_id"].toString())
--- a/tests/smoke/kotlin/src/test/resources/valid_manifests.json
+++ b/tests/smoke/kotlin/src/test/resources/valid_manifests.json
@@ -38,7 +38,7 @@
         "name": "ncird",
         "path_template": {
           "LOCAL": "/upload/ncird/{uid}",
-          "DEV": "/dex-dev/{dataStream}-{route}/{year}/{month}/{day}/{filename}",
+          "DEV": "/dex/{dataStream}-{route}/{year}/{month}/{day}/{filename}",
           "TEST": "/dex-test/{dataStream}-{route}/{year}/{month}/{day}/{filename}",
           "STAGE": "/dex/{dataStream}-{route}/{year}/{month}/{day}/{filename}"
         }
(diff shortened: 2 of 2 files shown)


cyber-decker:
Not all tests will pass right now.  These are the following known issues.

1) EICR/EDHI FileCopy tests failing due those delivery services being down.
2) Healthcheck fail in DEV/TEST due to EICR and EDHI services being DOWN. 
3) ProcStat test failed in DEV.  Passed in TEST.  Known issue.

tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
```diff
@@ -60,7 +60,8 @@ class FileCopy {
         val currentDateTime = ZonedDateTime.now(TimeZone.getTimeZone("GMT").toZoneId())
         uploadInfo.deliveries?.forEach { delivery ->
             Assert.assertEquals(delivery.status, "SUCCESS") // remove the assertion above?
-            val actualLocation = URLDecoder.decode(delivery.location, StandardCharsets.UTF_8.toString())
+            val encodedLocation = delivery.location.replace("+", "%2B")
```
cfarmer-fearless, line 63: Wondering if it's clearer to re-encode the whole thing and then decode.  Def add a comment why this is needed though.

cyber-decker: The URL is already partly encoded which is the problem.  The date piece is encoded in there already and that's how we get it.  I can add a comment here to mention it, though.

If anything this might be a server side problem in that it's inconsistently encoding the URL.  Should we put in a bug maybe to deal with this server side?

tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
```diff
@@ -59,8 +59,9 @@ class FileCopy {
 
         val currentDateTime = ZonedDateTime.now(TimeZone.getTimeZone("GMT").toZoneId())
         uploadInfo.deliveries?.forEach { delivery ->
-            Assert.assertEquals(delivery.status, "SUCCESS") // remove the assertion above?
-            val actualLocation = URLDecoder.decode(delivery.location, StandardCharsets.UTF_8.toString())
+            Assert.assertEquals(delivery.status, "SUCCESS") 
+            val encodedLocation = delivery.location.replace("+", "%2B")  // encode the + symbol in the location to properly decode it below
```
whytheplatypus, line 63: to make it clear what's going on (encoding and then decoding) Use a URLEncoder on location here.

cyber-decker: I don't think that should/could be done here.  This line in the test is already kind of a hack because the location we are getting back is already encoded but not fully encoded.   The + symbol is NOT getting encoded from the server side.  Now, I could pass delivery.location through URLEncoder.encode, but this is problematic because location contains a path that is already partially URL encoded, for example, one location as it comes back from the server will look like this: `https://ncirdndspdevsa.blob.core.windows.net/dex/generic-immunization-other%2F2024%2F11%2F12%2Fdex-smoke-test` Pushing this through encode gives us: `https%3A%2F%2Fncirdndspdevsa.blob.core.windows.net%2Fdex%2Fgeneric-immunization-other%252F2024%252F11%252F12%252Fdex-smoke-test` which is now a beast of a location, and NOT what we want to be trying to build an expected location against.   

More and more, this is feeling like a server side bug and we will have to address it there so that this line to replace the + here isn't necessary.

cyber-decker:
Okay.  So after discussion w/ @cfarmer-fearless I think we're going to keep the hacky replace in there for now and we should revisit how we do more consistent encoding on the info endpoint and throughout the application.   We should make sure everything in the path is encoded consistently (making it not human readable) or make it all decoded consistently (being more human readable).  I have opened issue PHDO-213 in JIRA for this.

tests/smoke/kotlin/src/test/kotlin/FileCopy.kt
cyber-decker: @whytheplatypus see comment below. I think we can leave this hack in there for now, but we should deal with the encoding/decoding on the server side to make it consistent/better.

whytheplatypus (approved):

PR #568, opened 2024-11-07
Run fortefy scan on PRs only

whytheplatypus:


cyber-decker (commented):
Spoke w/ @whytheplatypus and while these changes seem simple, there might be opportunity to fix how the fortify scans run during merge so that it works and doesn't fail.  Might be worth repurposing this effort here to fix that if possible, but this change here is a backup.  

Leaving room open here for now to see if fixing main merge with fority is possible.

whytheplatypus:
planning to bring this in as is until we can identify the best way to consistently get the correct ref in both the push to `main` trigger and the `pull_request` trigger ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context))

cyber-decker (approved):
This looks good as it is.  Simple change, makes sense here after discussing w/ @whytheplatypus

PR #569, opened 2024-11-07
UPLOAD-1825/delete-workflow

thetif:
Workflow to delete Playwright reports from the gh-pages branch after the branch is deleted


candace-campbell (approved):
LGTM

PR #574, opened 2024-11-15
Enable local integration of phdo components

whytheplatypus:
- Runs pstatus configured to use localstack sns/sqs
- assumes the repositories are next to eachother in the filesystem

### DIFF
--- a/upload-server/internal/event/aws.go
+++ b/upload-server/internal/event/aws.go
@@ -280,7 +280,7 @@ func (s *SQSSubscriber[T]) getBatch(ctx context.Context, max int) ([]types.Messa
 		WaitTimeSeconds:     1,
 		VisibilityTimeout:   int32(DefaultMessageVisibility),
 	})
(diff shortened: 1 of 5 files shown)


upload-server/internal/event/aws.go
```diff
@@ -280,7 +280,7 @@ func (s *SQSSubscriber[T]) getBatch(ctx context.Context, max int) ([]types.Messa
 		WaitTimeSeconds:     1,
 		VisibilityTimeout:   int32(DefaultMessageVisibility),
 	})
-	slog.Info("Got response", "rsp", rsp)
+	slog.Debug("Got response", "rsp", rsp)
```
cfarmer-fearless, line 283: Debug message can be more specific like "sqs receive message response"

cfarmer-fearless (changes requested):

cfarmer-fearless (approved):

PR #580, opened 2024-12-04
Use uploadID prefix (s3 only) for reports

whytheplatypus:
Before this reports in the aws environment were split between those with the TUS upload ID and the TUS ID + s3 multipart

### DIFF
--- a/upload-server/pkg/reports/builder.go
+++ b/upload-server/pkg/reports/builder.go
@@ -2,6 +2,7 @@ package reports
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
@@ -145,6 +146,7 @@ type Builder[T any] interface {
 }
 
 func NewBuilder[T any](version string, action string, uploadId string, dispType string) Builder[T] {
+	uploadId, _, _ = strings.Cut(uploadId, "+")
 	return &ReportBuilder[T]{
 		Version:         version,
 		Action:          action,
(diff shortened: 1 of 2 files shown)


upload-server/pkg/reports/builder.go
```diff
@@ -188,7 +191,6 @@ func (b *ReportBuilder[T]) SetAction(s string) Builder[T] {
 }
 
 func (b *ReportBuilder[T]) SetUploadId(id string) Builder[T] {
-	b.UploadId = id
```
cfarmer-fearless, line 191: Why remove this?

cfarmer-fearless (changes requested):
Seems like we're adding to the bandaid solution for handling the S3 ID that tus creates.  I think it would be better to do these string manipulation operations at a higher level than down in the report builder.  We can punt for now since we're trying to get the AWS migration complete, but will add to the tech debt.  We can discuss more and I can approve once my other above comment is addressed.

upload-server/pkg/reports/builder.go
whytheplatypus: wow sorry not sure how that got in

cfarmer-fearless (approved):

upload-server/internal/stores3/store.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		id = info.ID
+	}
+	u, err := s.Store.GetUpload(ctx, id)
+	return &S3StoreUpload{
+		u,
+	}, err
+}
+
+func (s S3Store) AsTerminatableUpload(upload handler.Upload) handler.TerminatableUpload {
```
cfarmer-fearless, line 73: Use either all pointer or non-pointer receivers for consistency.

cfarmer-fearless (changes requested):

cfarmer-fearless (approved):

PR #585, opened 2025-01-07
configure the timeout for considering a test failed

whytheplatypus:


tests/bad-uploader/config.go
```diff
@@ -222,6 +223,7 @@ func init() {
 	flag.DurationVar(&duration, "duration", 0, "the duration to run load for.")
 	flag.StringVar(&patchURL, "patch-url", "", "Override the base url to use to upload the file itself after upload creation.")
 	flag.BoolVar(&cases.random, "random", false, "Randomly select the next test case to run, only affects anything if multiple test cases are used.")
+	flag.DurationVar(&timeout, "timeout", time.Duration(30*time.Second), "The amount of time to wait for a successful delivery before calling it failed.")
```
candace-campbell, line 226: does this set the default to 30 seconds?

candace-campbell (approved):

tests/bad-uploader/config.go
whytheplatypus: yup!

cyber-decker: This does not seem to override what is in a test case file and I think it should.   If I set a timeout in a test case and a timeout here, the test case will continue with the value it has set at that level.  I think a value that is set with the timeout flag should override a value in a test case because it is more visible and apparent.  

However, the default should NOT override what is in a test case.  Let the test case do what it suggests and the default of 30 seconds should not override the test case.

cyber-decker (changes requested):

tests/bad-uploader/config.go
```diff
@@ -239,7 +241,7 @@ func init() {
 			Chunk:                   chunk,
 			Size:                    size,
 			Manifest:                manifest,
-			TimeLimit:               Duration(10 * time.Second),
+			TimeLimit:               Duration(timeout),
```
cyber-decker, line 244: I'm still running into some issues here.   I've tried running some tests with the timeout and it's not behaving as I expect.   So I'm running the following command

`go run ./... -load=1 -url=https://upload.phdo-eks-test.cdc.gov/files -patch-url=https://upload.phdo-eks-test.cdc.gov/files -case-file=".\tests\default\test-default.json" -v -timeout=1s`

the `test-default.json` I'm using is coming from the [test.default](https://github.com/CDCgov/data-exchange-upload/blob/main/tests/bad-uploader/tests/test-default.json) in main.   This test case file has a 11 second timeout built in.  

When running with this command with a timeout in the command as a flag, it still passes.  I thought that maybe this is because 1 second is too long of a timeout, but when setting a 1 second timeout inside the test case file it fails, which is really strange.  

1s timeout in the `time_limit` set in the test case will fail validation.  1s timeout in the -timeout flag will not fail, even if I remove the `time_limit` in the test case. I'm not sure the timeout is working as we expect?   Something is behaving kind of weird here. 

Not sure if there's a better way to validate this since the validation would need to slow down requests coming in and out of the system, but a 1s timeout consistently fails when it's set in the test case, but not as a timeout flag.

cyber-decker (changes requested):

tests/bad-uploader/config.go
cfarmer-fearless: I wasn't able to recreate this issue.  Timeout seems to be working as expected.

PR #586, opened 2025-01-08
Adding test files for bad uploader

cyber-decker:
Added JSON test cases for bad uploader.

- default cases are for dextesting/testevent1
- folder exists for actual data streams, one test case per data stream


candace-campbell (approved):
LGTM

PR #602, opened 2025-02-18
Phdo 392/sams auth

cfarmer-fearless:
Adds JWT user session validation and integration with existing OAuth middleware.  This functionality requires the `OAUTH_ENABLED` AND `OAUTH_ISSUER_URL` environment variables to be set.  Currently, we support SAMS oauth, so the issuer URL should be the SAMS issuer URLs: `https://apigw-stg.cdc.gov:8443` for stg and `https://apigw.cdc.gov:8443` for prd.

Note that this patch does NOT add support for OAuth password grant or auth code.  It assumes the end user already has their JWT from SAMS, and simply validates it either in the Authorization header, or cookies.

Token validation flow for API routes:
- Client sends request to Upload API server protected route (/files or /info)
- Server middleware reads bearer token from Authorization header
- Middleware performs JWT validation on said token
- Request is forwarded to the next handler if JWT is valid
- 401 Unauthorized is returned otherwise

Login flow for the UI:
- Browser client navigates to the UI console
- UI middleware reads and validates JWT from browser cookies under the name of `phdo_auth_token`
- If valid, request is forwarded to next hander, which is usually the UI handler that renders the page
- Otherwise, browser is redirected to `/login`, which is an unprotected route that simply renders the login page
- To login, end user inputs a JWT into the login form
- If valid, user is redirected to the protected route they were origionally requesting
- Otherwise, they are redirected back to the login page

TODO:
- Query param for redirection route (done)
- Unit tests for user session middleware (done)
- Cleanup (done)


upload-server/internal/middleware/authverification.go
```diff
@@ -1,15 +1,18 @@
...
+	"net/url"
 	"strings"
 
-	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/cdcgov/data-exchange-upload/upload-server/internal/oauth"
 )
 
+const UserSessionCookieName = "phdo_auth_token"
+
+var protectedUIRoutes = [...]string{"/", "/manifest", "/upload", "/status"}
```
cyber-decker, line 14: Should this be a configuration option somewhere outside of here?  Hardcoding might make this tricky to update if there are new endpoints that should be protected.

cfarmer-fearless: Yeah good call, I am not a fan of the hardcoding either.  Perhaps if I can make things dynamic based on the UI routes that the middleware is wrapping, that would suffice and IMO be better than a configuration option that could also be error prone

cfarmer-fearless: Since we'll be revisiting this in the security iteration, can we pass on this for now so these changes can get merged to the baseline?

cyber-decker: Not something we have to change here.  We can handle this later!

PR #604, opened 2025-02-21
PHDO-554/SSC Auth Findings

cfarmer-fearless:


tests/smoke/playwright/test-auth/upload-api-auth.spec.ts
```diff
@@ -158,30 +178,6 @@ test.describe('Upload API Auth', () => {
         await expect(page.getByText("Welcome to PHDO Upload Login")).toBeVisible()
     })
 
-    test('can log out after logging in', async ({ browser }) => {
```
cyber-decker, line 161: Saw this functionality was added to the end of the e2e test below, but might be worth having this be a small tests that logs in and logs out so we know this functionality works in the simplest way possible, no obstacles or barriers to proving this feature.  

this can be swapped to a regular form entry login rather than cookie.

cyber-decker (changes requested):
One comment for change, but this looks great.

cyber-decker (approved):

PR #607, opened 2025-02-25
Fixup/auth middleware and tus cors

cfarmer-fearless:
Patch to improve auth capabilities.  Namely making sure server doesn't fall over if issuer not available, better error handling of middleware init, and improve cors config for tusd to not break existing browser clients.  In addition, add healthcheck for the oidc provider


cyber-decker (approved):
Did some pairing and this is looking good.  Tested locally and against new playwright tests coming soon.  Tested with a mock oauth server as well.  Looks solid!  Good improvements/fixes here.

PR #608, opened 2025-02-25
Playwright auth E2E tests and mock auth server

cyber-decker:
Adding auth tests for when auth is enabled.  These tests will be run from a separate directory and not run with all the other tests since this feature may not be on all the time.

To run: 

1. From `/tests/smoke/playwright/resources/mock-auth-server` run `node server.js`
2. `npx playwright test -c playwright.auth.config.ts`


tests/smoke/playwright/resources/mock-auth-server/server.js
```diff
@@ -0,0 +1,133 @@
...
+    console.log("Public Key: ", publicKey)
+    console.log("Generated JWKS:", jwks);
+}
+
+generateKeys(); // Generate RSA keys at startup
+
+// OpenID Configuration Endpoint
+app.get('/.well-known/openid-configuration', (req, res) => {
+    res.json({
+        issuer: "http://localhost:3000",
```
cfarmer-fearless, line 72: Can this host and port number come from environment variables and default to localhost:3000?  I'll set us up nicely when this is run in a docker network.

tests/smoke/playwright/test-auth/upload-api-auth.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+        },
+        {
+            pageName: "login page",
+            path: "/login",
+            expectedRedirect: ""
+        },
+        {
+            pageName: "manifest page",
+            path: "/manifest?data_stream_id=dextesting&data_stream_route=testevent1",
+            expectedRedirect: "?redirect=/manifest?data_stream_id%3Ddextesting%26data_stream_route%3Dtestevent1"
```
cfarmer-fearless, line 20: This redirect value will soon come from the session cookie, not the url query params, just FYI

tests/smoke/playwright/test-auth/upload-api-auth.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    test('cannot log in with invalid scopes on token through the login page', async ({ page }) => {
+        await page.goto('/')  
+        await page.getByRole('textbox', { name: "Authentication Token *" }).fill(scopesToken)
+        await page.getByRole('button', { name: 'Login' }).click()
+        await expect(page.url()).toContain("/login")
+        await expect(page.getByText("Welcome to PHDO Upload Login")).toBeVisible()
+    })
+    
+    test('logs in with a valid cookie token', async ({ browser }) => {
```
cfarmer-fearless, line 93: I believe these tests that create cookies will soon become "can't login with forged cookie", and the happy path tests will be "failed login attempt creates user session cookie with redirect" and "successful login creates user session with jwt"

cfarmer-fearless (commented):
Def a solid foundation here.  Let's polish this up a bit and get it into main, then do some refactoring in my other PR for security findings.

tests/smoke/playwright/resources/mock-auth-server/server.js
cyber-decker: Yes, absolutely.  That was the plan to figure out how to do that next and also how to get this to the tests too.  I'll start looking at this very soon.

tests/smoke/playwright/test-auth/upload-api-auth.spec.ts
cyber-decker: Going to keep this in for now, and I think we may be able to change the logic of the test to try and inspect for that elsewhere.

tests/smoke/playwright/test-auth/upload-api-auth.spec.ts
cyber-decker: Sounds good, I think we can add that functionality here to validate the session as we build that functionality in.

tests/smoke/playwright/resources/mock-auth-server/server.js
```diff
@@ -0,0 +1,134 @@
+require('dotenv').config();
+const express = require('express');
+const cors = require('cors');
+const crypto = require('crypto');
+const jwt = require('jsonwebtoken');
+const base64url = require('base64url');
+
+const app = express();
+const PORT = process.env.MOCK_AUTH_PORT || 3000;
+const BASE_URL = `${process.env.MOCK_AUTH_BASE_URL}:${PORT}` || `http://localhost:${PORT}`;
```
cfarmer-fearless, line 10: Need to fix up this default for BASE_URL.  The value before the || will always be non-empty string since you default the port number, so the expected default of `http://localhost:{PORT}` won't take and BASE_URL will actually be `undefined:{PORT}` if MOCK_AUTH_BASE_URL is not set.

cfarmer-fearless (changes requested):

tests/smoke/playwright/resources/mock-auth-server/server.js
cyber-decker: Fixed up, good catch here.  This was a goofy way to do it.  Handled it more appropriately.

cfarmer-fearless (approved):

PR #612, opened 2025-03-07
Phdo 550/auth docs

cfarmer-fearless:
Adds documentation to the repo's README files to describe the new auth aspect of the system.  Specifically includes:
- New environment variables and configurations
- New behavior of both server and UI

This PR also includes removal of the EICR configuration documentation as that datastream has been offboarded.


cyber-decker (approved):
This looks great.  The documentation is clear and it outlines the configurations that are there.  Makes a lot of sense.

PR #621, opened 2025-03-24
PHDO-631/upload-speed-histogram-buckets

cfarmer-fearless:
Quick patch to improve the accuracy of the median upload speed calculation in prometheus by using finer grained buckets for lower values.  At some point, this should probably be a configurable thing.

This also changes the hook that runs metrics for post processing to the postfinish hook to not interfere with the synchronous response that the upload client gets back when an upload is finished.

Also includes JSON for the grafana dashboard that matches the live environments.

### DIFF
--- a/upload-server/configs/local/grafana/provisioning/dashboards/dex.json
+++ b/upload-server/configs/local/grafana/provisioning/dashboards/dex.json
@@ -15,16 +15,286 @@
       }
     ]
   },
+  "description": "Visualization of various metrics sourced by PHDO microservices to give an idea of overall health and performance of the system.",
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
+  "id": 122,
   "links": [],
   "panels": [
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "PBFA97CFB590B2093"
+        "uid": "prometheus-datasource"
       },
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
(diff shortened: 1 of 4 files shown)


cfarmer-fearless:
Merged into PHDO-630

upload-server/configs/local/grafana/provisioning/dashboards/dex.json
```diff
@@ -80,87 +50,55 @@
...
       "targets": [
         {
-          "datasource": {
-            "type": "prometheus",
-            "uid": "PBFA97CFB590B2093"
-          },
           "disableTextWrap": false,
           "editorMode": "builder",
-          "expr": "sum by(data_stream_id, data_stream_route, sender_id) (rate(upload_manifest_count[$__rate_interval]))",
+          "expr": "sum by(job) (dex_server_active_uploads)",
```
mkrystof, line 81: Under the "panels.options" section of the json you can add an "xaxis.title".  Recommend adding an xaxis title for all of the panels or else the "expr" gets used as the x-axis label.  Specifying one ourselves makes it a little more human readable and easier to understand what it represents.

mkrystof (changes requested):
Just the one suggestion about the x-axis labels.

upload-server/configs/local/grafana/provisioning/dashboards/dex.json
cfarmer-fearless: I believe I'm already setting the titles already.  All panels and series labels have custom titles that are more human readable.

mkrystof: Gotcha, I didn't look at your latest panels - just recalling from what I saw when you demo'd it.

PR #624, opened 2025-04-08
Phdo 677/prom metrics pending deliveries

cfarmer-fearless:
Adds prometheus metrics for counting events that the upload server publishes and subscribes to, as well as periodically sets a gauge for the current number of messages in a queue.  These counts can be used to determine number of files that have and have not yet been delivered or have failed to be handled.

Here are all the new metrics that are being emitted:
image
**Note that the counts for the report-type events are doubled because the server emits them to the file system as well as the remote queue that PS API consumes.**

**Note that for testing of the queue gauge locally, it is likely necessary to add a delay in the file ready processing because messages are handled faster than the default sampling rate for the queue size.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.7.2] 2025-3-31
+- Prometheus histogram for upload speed in bytes per second.
+
 ## [2.7.1] 2025-3-6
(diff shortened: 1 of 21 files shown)


cyber-decker (approved):
This looks good and makes sense to me.  I was able to get this running locally and saw the metrics populating as expected.

I was able to get the gauge setup in grafana for the event queue.  Did have to add a sleep in the FileProcessReadyEvent but that did the trick and was able to see this populate and did see the gauge in grafana as well when I pointed one to this metric.  

This metric seems very time sensitive, and maybe it's a good indicator if we do see it that there are a lot of things stuck in the queue for an extended period of time.  I'd be curious to see how this behaves in production environments under a normal load or a moderate load to get an idea of what this metric should tell us when things are good or bad.

PR #630, opened 2025-04-14
Phdo 516 upload id logger

achoi-mechanicode:
Propagate uploadIds in log messages automatically by implementing the following approach which replaces the default slog with a logger that is updated with the uploadId and stored/retrieved from the context:

Per CoPilot:
Store the Logger: Use context.WithValue to store the logger in the context.
Retrieve the Logger: Use a helper function like getLogger to retrieve the logger from the context.
Fallback to Default: If the logger is not found in the context, fall back to the default logger.
This approach ensures that the logger is thread-safe and tied to the context, making it easy to propagate logging information throughout the application.


upload-server/internal/metadata/metadata.go
```diff
@@ -103,16 +105,19 @@ func (v *SenderManifestVerification) Verify(event handler.HookEvent, resp hooks.
...
-	slog.Info("checking the sender manifest", "manifest", manifest, "uploadId", tuid)
-
 	if resp.ChangeFileInfo.ID != "" {
 		tuid = resp.ChangeFileInfo.ID
 	}
 	if tuid == "" {
 		return resp, errors.New("no Upload ID defined")
 	}
 
+	ctx := sloger.SetUploadId(event.Context, tuid)
```
cfarmer-fearless, line 115: For the tus hooks, I think it would be better to set up the logger as a hook handler function that gets called when the hook in invoked, as opposed to sprinkling calls to `sloger.SetUploadId` in the handler functions.  The only catch is the function needs to be invoked after the upload ID gets set, which happens here https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/internal/metadata/metadata.go#L232

upload-server/internal/event/publish.go
```diff
@@ -4,25 +4,26 @@ import (
...
 		for range MaxRetries {
 			if err := publisher.Publish(ctx, e); err != nil {
-				slog.Error("Failed to publish", "event", e, "publisher", publisher, "err", err)
+				logger.Error("Failed to publish", "event", e, "publisher", publisher, "err", err)
 				errs = fmt.Errorf("Failed to publish event %s %w", e.Identifier(), err)
 				continue
 			}
 			metrics.EventsCounter.With(prometheus.Labels{metrics.Labels.EventType: e.Type(), metrics.Labels.EventOp: "publish"}).Inc()
-			slog.Info("published event", "event", e, "uploadId", e.Identifier())
+			logger.Info("published event", "event", e, "uploadId", e.Identifier())
```
cfarmer-fearless, line 26: Shouldn't need to include the upload ID in this log.  The `logger` should have it in its context.

cfarmer-fearless (changes requested):
This approach looks good!  Logging still needs to be enhanced for the subscribe side of things, like for file delivery so we can see the upload ID in the whole critical path for an upload.  I also think there's a bit of code cleanup.  Great start here though.

upload-server/internal/metadata/metadata.go
cfarmer-fearless: Most hook handlers get registered here https://github.com/CDCgov/data-exchange-upload/blob/main/upload-server/cmd/cli/hooks.go#L46

cfarmer-fearless (approved):

PR #633, opened 2025-04-22
PHDO-719/tracing instrumentation

cfarmer-fearless:
This patch adds instrumentation for Open Telemetry tracing.  The upload server emits tracing data via http to a remote exporter.  For PHDO, this will be Tempo.  The exporter in turn stores the tracing data and exposes it in our Grafana instance for querying and visualization.

This instrumentation can be tested locally as well.  Simply run the monitoring stack locally with the following command: `podman-compose -f docker-compose.yml -f docker-compose.monitoring.yml up -d`.  Next, upload a file.  Then, take the MD5 hash of the upload ID for that file.  Finally, open grafana at `http://localhost:3000`.  In the explore tab, select tempo from the data sources drop down.  In the TraceQL input field, paste in the MD5 hash.  You should see traces show up that show the file being uploaded, and then delivered.

image

### DIFF
--- a/upload-server/cmd/main.go
+++ b/upload-server/cmd/main.go
@@ -54,6 +54,12 @@ func main() {
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	var mainWaitGroup sync.WaitGroup
 
+	tpShutdown, err := cli.InitTracerProvider(ctx)
+	if err != nil {
+		slog.Error("error starting app, error starting tracing", "error", err)
+		os.Exit(appMainExitCode)
+	}
+
 	appConfig, err := appconfig.ParseConfig(ctx)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cyber-decker (approved):
This is looking good.  Was able to get this up and working just fine.  This let me see some really interesting differences in how the web UI and the api for upload work differently by collecting all the requests in one place to see them better.

Wished there was a better way rather than just MD5 hashing the upload ID to get the trace, but this is effective and does the trick.  What if we include the md5 hash in the upload info at some point as well?  Not a thing we have to do now, but maybe sometime down the line.

This looks good, was able to test it with the UI, bad uploader, with single requests and both small and large file loads.   Did the trick in all cases!  Looks awesome.

PR #635, opened 2025-04-25
Phdo 729/bugfix delivery failed metric

cfarmer-fearless:
Patch to fix the failed delivery counter not being tracked correctly.  This was due to an inconsistency in the label name that was being incremented, and the one that Grafana was looking for.


mkrystof (approved):
The changes look pretty minor and I don't see any issues or have any concerns.  At some point I'd love to test these dashboard locally as I know you and others have gone through that exercise already of setting up a local env for that.

PR #637, opened 2025-04-28
Phdo 683/dashboard time since last delivery

cfarmer-fearless:
This patch adds a panel to the snapshot dashboard to show a table for time since last successful delivery for each unique delivery target.  This required a new prometheus rule to be added to pre-calculate the timestamp of the increase in the delivery complete counter.

image


cyber-decker (approved):
This looks great.  There is a known issue where the very first metric reporting, similar to the Last Successful Upload also doesn't show up, but after that first one populates, everything is looking good.  Was able to get this working with multiple data streams, and it looks good.

PR #641, opened 2025-05-01
PHDO-693 Active Upload Alternative Metric

cfarmer-fearless:
Problem: The metric `dex_server_active_uploads` gets frequently out of sync in distributed scenarios where there are multiple replicas of this service.  In these situations, race conditions happen that make the counter report inaccurate values, such as negative values.  In addition, this metric does not take into account uploads that have been abandoned.  In the grafana dashboard gauge, this gives the false impression that an upload is in progress, when really it will never finish.  Overall, this is a consequence of Tusd as it doesn't keep track of uploads that haven't received a chunk after a certain amount of time, or have otherwise gone stale.

Mitigation: To mitigate this, we use a different promQL query that relies on two counters that are already being emitted by Tusd: `tusd_uploads_created` and `tusd_uploads_finished`.  These are counters that increment when an upload initiates and when the last chunk is received.  Naturally, taking the difference of these counts gives us uploads in progress:
```
sum(tusd_upload_created) - sum(tusd_uploads_finished)
```
This approach fixes the first problem of race conditions sense these metrics are counts instead of gauges.  However, it is still is susceptible to the second problem of counting abandoned uploads.  To account for this, we'll use the `increase` function where we can specify a time window for the counts that we want to track.  We can set this time window to one that matches our SLA of 1 hour.  With that, prometheus will not count increases that are older than 1 hour.  The counters themselves will still show a greater count for uploads started, but prometheus will ignore those increases after an hour.  This brings us to our final query:
```
round(sum(increase(tusd_uploads_created[1h])) - sum(increase(tusd_uploads_finished[1h])))
```

### DIFF
--- a/upload-server/configs/local/grafana/provisioning/dashboards/dex.json
+++ b/upload-server/configs/local/grafana/provisioning/dashboards/dex.json
@@ -40,12 +40,15 @@
         "type": "prometheus",
         "uid": "prometheus-datasource"
       },
+      "description": "Gauge to show the number of files that have been started over a time window of 1 hour.",
       "fieldConfig": {
         "defaults": {
           "color": {
             "mode": "thresholds"
           },
           "mappings": [],
+          "max": 100,
+          "min": 0,
           "thresholds": {
             "mode": "absolute",
             "steps": [
@@ -53,8 +56,16 @@
                 "color": "green"
               },
               {
+                "color": "#EAB839",
+                "value": 50
+              },
+              {
                 "color": "red",
                 "value": 80
+              },
+              {
+                "color": "dark-red",
+                "value": 100
               }
             ]
           }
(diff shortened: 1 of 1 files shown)


cyber-decker (approved):
I think this is looking good.  

Did some testing around this and the feature seems to be working okay and reporting back the active uploads mostly in an appropriate way.  This will be a better way to track this going forward.

I did notice ONE behavior that is a little weird, but understandable due to the way this is calculated and I don't know if there's a better way to handle this.  Starting an upload (even if no bytes are sent) that this counts as active, as expected.  However, if the (configured) 1 hour window lapses, and the upload is then resumed and completed, this actually counts against the total and will end up as negative when done.   It's small, and will only last through the next window and will settle back down to zero, but I think this is a very minor thing that we can't really get around.

All in all, this looks great and is a nice metric to have.

